PR #49, opened 2020-04-27
Adding Support for configurable Pod Readiness

CharlyF:
Adding extra logic to the controller to avoid scenarios where the Target would be scaled with Pending pods that would never go in a Running Phase [1].
With the current behaviour, all pods in pending or running phases would be accounted for in the multiplier to compute the recommended number of replicas.
This could yield an exponential growth majored by the ScaleupFactor (which is not desirable).
We believe that while this is not a bug (as it is an expected behavior), it is a limitation that ought to be removed.

We are thus extending the logic of the External Metrics loop to support the `ReadinessDelaySeconds` parameter introduced when adding support for Resource Metrics. 
This introduces a new behaviour. Henceforth, by default, `Pending` pods will *not* be accounted for.
If you believe that pods in a Pending can be accounted for (if the images are large, or pods take a while to transition to a `Running` phase), specify this value so that you do not autoscale more than necessary.

## Note:

- This behavioural change can introduce an unexpected scaling event. If the metric used is within the bounds, but you have a number of pods that are Pending and considered expired. Proven that scaling is authorised (forbidden windows are cleared), the autoscaler might adjust the number of replicas to the number of `Running` ones.

- We are adding an Informer on pods to align with the signatures of the upstream controllers. This could result in a higher memory footprint in large clusters.

## TODO 

- [x] Uses the scale.Status.ReplicaCount instead of the  scale.Spec.ReplicaCount (which was inherently wrong)
- [x] Test new behaviour
- [x] Resource Metrics do not match this behaviour yet. Pending pods are not given a tolerance 
- [x] Add an informer to the controller
- [ ] Document the new behaviour (will be done in an upcoming PR)
- [x] Update RBAC (document the need to list/watch pods).
- [x] Last pass to review the code.


pkg/controller/watermarkpodautoscaler/replica_calculator.go
```diff
@@ -32,29 +33,42 @@ type ReplicaCalculation struct {
...
 // namespace, and the current replica count.
-func (c *ReplicaCalculator) GetExternalMetricReplicas(logger logr.Logger, currentReplicas int32, metric v1alpha1.MetricSpec, wpa *v1alpha1.WatermarkPodAutoscaler) (ReplicaCalculation, error) {
+func (c *ReplicaCalculator) GetExternalMetricReplicas(logger logr.Logger, target *autoscalingv1.Scale, metric v1alpha1.MetricSpec, wpa *v1alpha1.WatermarkPodAutoscaler) (ReplicaCalculation, error) {
+	lbl, err := labels.Parse(target.Status.Selector)
+	if err != nil {
+		log.Error(err, "Could not parse the labels of the target")
+	}
+	currentReadyReplicas, err := c.getReadyPodsCount(metav1.NamespaceAll, lbl, time.Duration(wpa.Spec.ReadinessDelaySeconds)*time.Second)
+	if err != nil {
+		return ReplicaCalculation{0, 0, time.Time{}}, fmt.Errorf("unable to get the number of ready pods across all namespaces for %v: %s", lbl, err.Error())
```
clamoriniere, line 65: it is equivalent to: `ReplicaCalculation{}`

pkg/controller/watermarkpodautoscaler/replica_calculator.go
```diff
@@ -79,25 +93,20 @@ func (c *ReplicaCalculator) GetExternalMetricReplicas(logger logr.Logger, curren
...
-	}
+	logger.V(4).Info("Metrics from the External Metrics Provider", "metrics", metrics)
 
 	var sum int64
 	for _, val := range metrics {
 		sum += val
 	}
-
-	replicaCount, utilizationQuantity := getReplicaCount(logger, currentReplicas, wpa, metricName, averaged, sum, metric.External.LowWatermark, metric.External.HighWatermark)
+	adjustedUsage := float64(sum) / averaged
```
clamoriniere, line 102: could you add comment to this operation?
I don't understand what means adjusted here. Is average can be equal to 0.0 ?

pkg/controller/watermarkpodautoscaler/watermarkpodautoscaler_controller.go
```diff
@@ -57,6 +61,22 @@ var (
...
 )
 
+func initializePodInformer(clientConfig *rest.Config) listerv1.PodLister {
+	a := ctrl.SimpleControllerClientBuilder{ClientConfig: clientConfig}
+	versionedClient := a.ClientOrDie("watermark-pod-autoscaler-shared-informer")
+	// Only resync every 5 minutes.
+	// TODO Consider exposing configuration of the resync for the pod informer.
+	sharedInf := informers.NewSharedInformerFactory(versionedClient, 300*time.Second)
+
+	// TODO add better context handling
```
clamoriniere, line 71: did you plan to implement it in the PR?

pkg/controller/watermarkpodautoscaler/watermarkpodautoscaler_controller.go
```diff
@@ -65,9 +85,12 @@ func newReconciler(mgr manager.Manager) (reconcile.Reconciler, error) {
 		nil,
 		external_metrics.NewForConfigOrDie(clientConfig),
 	)
+
+	podLister := initializePodInformer(clientConfig)
```
clamoriniere, line 89: to be about to pass a stop channel to the podLister. I see two options:
* You can add add a `context.Context`(or just a `stop chan`) in the `ReconcileWatermarkPodAutoscaler` struct. the `ReconcileWatermarkPodAutoscaler.Start(...)` receive a stop channel. so you can cascade after the stop event to the internal channel.
* you can initialize the pod informer only when you start the `ReconcileWatermarkPodAutoscaler`.

pkg/controller/watermarkpodautoscaler/replica_calculator.go
```diff
@@ -190,36 +204,67 @@ func getReplicaCount(logger logr.Logger, currentReplicas int32, wpa *v1alpha1.Wa
...
+	}
+
+	if len(podList) == 0 {
+		return 0, fmt.Errorf("no pods returned by selector while calculating replica count")
+	}
+
+	toleratedAsReadyPodCount := 0
+
+	for _, pod := range podList {
+		_, condition := getPodCondition(&pod.Status, corev1.PodReady)
```
clamoriniere, line 220: Should we check instead: `PodAvailable` instead? (at least ready once)

clamoriniere: finally I think your code is looking as I say for pod ready at least once. since you donc check the `condition.Status == true`

pkg/controller/watermarkpodautoscaler/replica_calculator.go
```diff
@@ -190,36 +204,67 @@ func getReplicaCount(logger logr.Logger, currentReplicas int32, wpa *v1alpha1.Wa
...
+
+func groupPods(logger logr.Logger, podList []*corev1.Pod, metrics metricsclient.PodMetricsInfo, resource corev1.ResourceName, delayOfInitialReadinessStatus time.Duration) (readyPods, ignoredPods sets.String) {
 	readyPods = sets.NewString()
 	ignoredPods = sets.NewString()
-	for _, pod := range podList.Items {
+	missing := sets.NewString()
+	for _, pod := range podList {
 		// Failed pods shouldn't produce metrics, but add to ignoredPods to be safe
-		if pod.Status.Phase == v1.PodFailed {
+		if pod.Status.Phase == corev1.PodFailed {
```
clamoriniere, line 242: could you check that evicted pod is also considered as `PodFailed`?

pkg/controller/watermarkpodautoscaler/watermarkpodautoscaler_controller.go
CharlyF: not in this, but in the future. yes.

pkg/controller/watermarkpodautoscaler/watermarkpodautoscaler_controller.go
CharlyF: I moved the channel outside of the initializePodInfomer method.
I will implement more logic to handle the cancelation in the reconcile loop in a subsequent PR.

pkg/controller/watermarkpodautoscaler/replica_calculator.go
CharlyF: PodAvailable is not available, I think PodReady here is already a good indicator as it means that the pod is able to service requests and should be added to the load balancing pools of all matching services.

pkg/controller/watermarkpodautoscaler/replica_calculator.go
CharlyF: Thanks for pointing this out.
Yes, evicted pods have a Status.Phase == Failed

```
kubectl get pods --all-namespaces | grep Evicted
kube-system                metrics-server-XXX-6gv6s     0/2     Evicted     0    213d

kubectl get po metrics-server-XXX-6gv6s -n kube-system -oyaml
[...]
status:
  message: 'The node was low on resource: imagefs.'
  phase: Failed
  reason: Evicted
  startTime: "2019-11-05T02:51:25Z"
```

PR #55, opened 2020-07-13
Fixing the pod informer selector 

CharlyF:
With the current logic, the selector used to retrieve the number of pods of the target to only use the healthy ones in the calculation of the recommendation, relies on the selector exposed by the `Scale` interface.
This selector is the one in the matchSelector in the target deploy (or sts). The issue is that this selector can be shared by multiple targets.
Therefore, you can end up in a scenario where the Target has 5 replicas and as we try to see how many out of these 5 are healthy using the pod informer, if multiple deployments have the same selector, we might see 18 pods.
The solution is to check the owner ref as well to make sure we only look at the pods that relate to the target we are autoscaling.

### DIFF
--- a/pkg/controller/watermarkpodautoscaler/replica_calculator.go
+++ b/pkg/controller/watermarkpodautoscaler/replica_calculator.go
@@ -104,7 +104,6 @@ func (c *ReplicaCalculator) GetExternalMetricReplicas(logger logr.Logger, target
 	// if the average algorithm is used, the metrics retrieved has to be divided by the number of available replicas.
 	adjustedUsage := float64(sum) / averaged
 	replicaCount, utilizationQuantity := getReplicaCount(logger, target.Status.Replicas, currentReadyReplicas, wpa, metricName, adjustedUsage, metric.External.LowWatermark, metric.External.HighWatermark)
-	log.Info("getRep", "currentReplicas", target.Status.Replicas, "currentReadyReplicas", currentReadyReplicas, "replicaCount", replicaCount)
 	return ReplicaCalculation{replicaCount, utilizationQuantity, timestamp}, nil
 }
 
@@ -245,7 +244,7 @@ func (c *ReplicaCalculator) getReadyPodsCount(target *autoscalingv1.Scale, selec
 }
 func checkOwnerRef(ownerRef []metav1.OwnerReference, targetName string) bool {
 	for _, o := range ownerRef {
-		if o.Kind != "ReplicaSet" {
+		if o.Kind != "ReplicaSet" && o.Kind != "StatefulSet" {
 			continue
 		}
 		if strings.HasPrefix(o.Name, targetName) {
--- a/pkg/controller/watermarkpodautoscaler/replica_calculator.go
+++ b/pkg/controller/watermarkpodautoscaler/replica_calculator.go
@@ -176,7 +176,7 @@ func (c *ReplicaCalculator) GetResourceReplicas(logger logr.Logger, target *auto
 	return ReplicaCalculation{replicaCount, utilizationQuantity, timestamp}, nil
 }
 
-func getReplicaCount(logger logr.Logger, currentReplicas, currentAvailReplicas int32, wpa *v1alpha1.WatermarkPodAutoscaler, name string, adjustedUsage float64, lowMark, highMark *resource.Quantity) (replicaCount int32, utilization int64) {
+func getReplicaCount(logger logr.Logger, currentReplicas, currentReadyReplicas int32, wpa *v1alpha1.WatermarkPodAutoscaler, name string, adjustedUsage float64, lowMark, highMark *resource.Quantity) (replicaCount int32, utilization int64) {
 	utilizationQuantity := resource.NewMilliQuantity(int64(adjustedUsage), resource.DecimalSI)
 	adjustedHM := float64(highMark.MilliValue()) + wpa.Spec.Tolerance*float64(highMark.MilliValue())
 	adjustedLM := float64(lowMark.MilliValue()) - wpa.Spec.Tolerance*float64(lowMark.MilliValue())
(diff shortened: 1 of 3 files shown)


pkg/controller/watermarkpodautoscaler/replica_calculator.go
```diff
@@ -232,17 +237,35 @@ func (c *ReplicaCalculator) getReadyPodsCount(namespace string, selector labels.
 			toleratedAsReadyPodCount++
 		}
 	}
+	log.Info("getReadyPodsCount", "full podList length", len(podList), "toleratedAsReadyPodCount", toleratedAsReadyPodCount, "incorrect target", incorrectTargetPods)
```
celenechang, line 240: similarly, perhaps `incorrect target` -> `incorrect target pod count`

celenechang: could it be worth logging the target name?

pkg/controller/watermarkpodautoscaler/replica_calculator.go
```diff
@@ -170,13 +172,12 @@ func (c *ReplicaCalculator) GetResourceReplicas(logger logr.Logger, target *auto
 	}
 	adjustedUsage := float64(sum) / averaged
 
-	replicaCount, utilizationQuantity := getReplicaCount(logger, target.Status.Replicas, wpa, string(resourceName), adjustedUsage, metric.Resource.LowWatermark, metric.Resource.HighWatermark)
+	replicaCount, utilizationQuantity := getReplicaCount(logger, target.Status.Replicas, int32(readyPodCount), wpa, string(resourceName), adjustedUsage, metric.Resource.LowWatermark, metric.Resource.HighWatermark)
 	return ReplicaCalculation{replicaCount, utilizationQuantity, timestamp}, nil
 }
 
-func getReplicaCount(logger logr.Logger, currentReplicas int32, wpa *v1alpha1.WatermarkPodAutoscaler, name string, adjustedUsage float64, lowMark, highMark *resource.Quantity) (replicaCount int32, utilization int64) {
+func getReplicaCount(logger logr.Logger, currentReplicas, currentAvailReplicas int32, wpa *v1alpha1.WatermarkPodAutoscaler, name string, adjustedUsage float64, lowMark, highMark *resource.Quantity) (replicaCount int32, utilization int64) {
```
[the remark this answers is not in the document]
CharlyF: yes you're right - I wanted to make the distinction between ready and available initially but did not go all the way

pkg/controller/watermarkpodautoscaler/replica_calculator.go
CharlyF: We log the target name a little before, the info I am trying to display is how many pods share the same label selector essentially.

CharlyF: will use `incorrectly targeted pods`

pkg/controller/watermarkpodautoscaler/replica_calculator.go
```diff
@@ -206,22 +207,26 @@ func getReplicaCount(logger logr.Logger, currentReplicas int32, wpa *v1alpha1.Wa
 	return replicaCount, utilizationQuantity.MilliValue()
 }
 
-func (c *ReplicaCalculator) getReadyPodsCount(namespace string, selector labels.Selector, readinessDelay time.Duration) (int32, error) {
-	podList, err := c.podLister.Pods(namespace).List(selector)
+func (c *ReplicaCalculator) getReadyPodsCount(target *autoscalingv1.Scale, selector labels.Selector, readinessDelay time.Duration) (int32, error) {
```
clamoriniere, line 210: should we rename this function `getAvailablePodsCount()`

pkg/controller/watermarkpodautoscaler/replica_calculator.go
```diff
@@ -93,7 +94,7 @@ func (c *ReplicaCalculator) GetExternalMetricReplicas(logger logr.Logger, target
 		value.Delete(prometheus.Labels{wpaNamePromLabel: wpa.Name, metricNamePromLabel: metricName})
 		return ReplicaCalculation{0, 0, time.Time{}}, fmt.Errorf("unable to get external metric %s/%s/%+v: %s", wpa.Namespace, metricName, selector, err)
 	}
-	logger.V(4).Info("Metrics from the External Metrics Provider", "metrics", metrics)
+	logger.Info("Metrics from the External Metrics Provider", "metrics", metrics)
```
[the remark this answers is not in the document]
CharlyF: yes, it's an important log (I parse it to extract the value to double check that the downstream value is the same as the one exposed by the cluster agent). We don't use the verbosity value elsewhere so there was no real incentive to to start the controller with a `-v 4` just for that.

pkg/controller/watermarkpodautoscaler/replica_calculator.go
```diff
@@ -102,7 +103,8 @@ func (c *ReplicaCalculator) GetExternalMetricReplicas(logger logr.Logger, target
 
 	// if the average algorithm is used, the metrics retrieved has to be divided by the number of available replicas.
 	adjustedUsage := float64(sum) / averaged
-	replicaCount, utilizationQuantity := getReplicaCount(logger, currentReadyReplicas, wpa, metricName, adjustedUsage, metric.External.LowWatermark, metric.External.HighWatermark)
+	replicaCount, utilizationQuantity := getReplicaCount(logger, target.Status.Replicas, currentReadyReplicas, wpa, metricName, adjustedUsage, metric.External.LowWatermark, metric.External.HighWatermark)
+	log.Info("getRep", "currentReplicas", target.Status.Replicas, "currentReadyReplicas", currentReadyReplicas, "replicaCount", replicaCount)
```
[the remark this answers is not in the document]
CharlyF: hum, that is an oversight, I will remove

pkg/controller/watermarkpodautoscaler/replica_calculator.go
```diff
@@ -232,17 +237,35 @@ func (c *ReplicaCalculator) getReadyPodsCount(namespace string, selector labels.
...
 	}
+	log.Info("getReadyPodsCount", "full podList length", len(podList), "toleratedAsReadyPodCount", toleratedAsReadyPodCount, "incorrect target", incorrectTargetPods)
 	if toleratedAsReadyPodCount == 0 {
 		return 0, fmt.Errorf("among the %d pods, none is ready. Skipping recommendation", len(podList))
 	}
 	return int32(toleratedAsReadyPodCount), nil
 }
+func checkOwnerRef(ownerRef []metav1.OwnerReference, targetName string) bool {
+	for _, o := range ownerRef {
+		if o.Kind != "ReplicaSet" {
```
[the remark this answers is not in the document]
CharlyF: will add

pkg/controller/watermarkpodautoscaler/replica_calculator.go
CharlyF: this returns the pods that are Ready though, not available

clamoriniere: should we use Available instead of ready?

Available means that the pods became ready at least once

pkg/controller/watermarkpodautoscaler/replica_calculator.go
```diff
@@ -232,17 +236,35 @@ func (c *ReplicaCalculator) getReadyPodsCount(namespace string, selector labels.
...
 	}
+	log.Info("getReadyPodsCount", "full podList length", len(podList), "toleratedAsReadyPodCount", toleratedAsReadyPodCount, "incorrectly targeted pods", incorrectTargetPodsCount)
 	if toleratedAsReadyPodCount == 0 {
 		return 0, fmt.Errorf("among the %d pods, none is ready. Skipping recommendation", len(podList))
 	}
 	return int32(toleratedAsReadyPodCount), nil
 }
+func checkOwnerRef(ownerRef []metav1.OwnerReference, targetName string) bool {
+	for _, o := range ownerRef {
+		if o.Kind != "ReplicaSet" && o.Kind != "StatefulSet" {
```
clamoriniere, line 247: you can add `ReplicationController` too

PR #57, opened 2020-08-06
add labels_info metric to support additional labels as tags

celenechang:
- Add an env var `DD_LABELS_AS_TAGS` that will add space-separated label keys as metric tags using label joins (with the wpa name label).

- Provides sample configuration in operator.yaml accordingly

To test:
- add desired labels in WPA CR
- update `DD_LABELS_AS_TAGS` and prometheus check configuration in operator.yaml
- confirm labels are present in prometheus metrics and tags are present in Datadog prometheus check output


clamoriniere:
Could you add an option in the [chart](https://github.com/DataDog/watermarkpodautoscaler/tree/master/chart/watermarkpodautoscaler), to ease the configuration of this new option

celenechang:
will make the chart update in a new PR

PR #61, opened 2020-08-25
Fix metrics cleanup

ahmed-mez:
### What does this PR do?

Pass the right metric contexts to make sure we clean up the metrics correclty

### Motivation

The following metrics were not deleted when deleting the corresponding Custom Resource

- wpa_controller_high_watermark
- wpa_controller_low_watermark
- wpa_controller_restricted_scaling
- wpa_controller_value


clamoriniere (approved):
good catch 💯

PR #47, opened 2020-02-26
Added troubleshooting notes for the cluster agent configuration

arapulido:
Add some troubleshooting notes for the Cluster Agent to enable the WPA listener and to be able to watch the WPA objects

PR #64, opened 2020-08-27
Add DCA helm chart instruction

ahmed-mez:
### What does this PR do?

Mention how to enable WPA in the DCA using helm

### Motivation

Complete this https://github.com/DataDog/watermarkpodautoscaler/pull/47

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,8 @@ The Cluster Agent doesn't run the WPA listener by default. To enable WPA in the 
 [...]
 ```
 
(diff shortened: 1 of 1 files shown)


clamoriniere (commented):
Should you add equivalent recommandation when it is used with the Operator?

ahmed-mez:
@clamoriniere we don't have a dedicated `wpa` field in the CRD [see here ](https://github.com/DataDog/datadog-operator/blob/master/pkg/apis/datadoghq/v1alpha1/datadogagent_types.go#L662)

PR #66, opened 2020-09-03
[wpa helm chart] Add missing rule for Pods in ClusterRole in helm chart

xornivore:
### What does this PR do?

Sync rules from `deploy/clusterrole.yaml` to `chart/watermarkpodautoscaler/templates/clusterrole.yaml`, note that we don't request `get` verb which is probably ok for what we need.

### Motivation

Working RBAC in helm chart.

### Additional Notes

### Describe your test plan

Ensure there are no logs similar to:
```
E0902 18:04:45.304836       1 reflector.go:123] pkg/controller/watermarkpodautoscaler/watermarkpodautoscaler_controller.go:73: Failed to list *v1.Pod: pods is forbidden: User "system:serviceaccount:default:wpacontroller-watermarkpodautoscaler" cannot list resource "pods" in API group "" at the cluster scope
```

### DIFF
--- a/chart/watermarkpodautoscaler/templates/clusterrole.yaml
+++ b/chart/watermarkpodautoscaler/templates/clusterrole.yaml
@@ -24,6 +24,13 @@ rules:
   - create
   - patch
 - apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - list
+  - watch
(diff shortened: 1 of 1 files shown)


clamoriniere (approved):
💯

PR #67, opened 2020-09-11
add labelsAsTags to chart template

celenechang:
### What does this PR do?

Updates chart for configuring labels as metric tags from #57 

### Motivation

Better user experience

### Describe your test plan

Test Datadog prometheus check output (i.e. the tags) for WPA metrics with and without `labelsAsTags` set in values.yaml

### DIFF
--- a/chart/watermarkpodautoscaler/values.yaml
+++ b/chart/watermarkpodautoscaler/values.yaml
@@ -25,9 +25,7 @@ rbac:
 # Configure the controller to watch all namespaces
 watchAllNamespaces: true
 # Labels to be included as metric tags
-# labelsAsTags:
-#   - label1
-#   - label2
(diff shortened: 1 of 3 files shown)


chart/watermarkpodautoscaler/values.yaml
```diff
@@ -24,6 +24,10 @@ rbac:
   create: true
 # Configure the controller to watch all namespaces
 watchAllNamespaces: true
+# Labels to be included as metric tags
+# labelsAsTags:
```
clamoriniere, line 28: we try to have all the fields define.   `{{- if .Values.labelsAsTags }}` still return false if any value is in the list.

```suggestion
labelsAsTags:
```

clamoriniere (approved):

PR #70, opened 2020-12-23
Upgrading to operator 0.19.0

CharlyF:
### What does this PR do?

- Bumping the operator.
- Main change is on the supported types, from strings to anyOf.
-  Enforce tolerance and scaling windows values in the code (can't use kubebuilder's features).
- Improve logging.

### Motivation

step update to get the operator 1.0.0.

### Describe your test plan

Update controller with existing WPAs and create new ones, make sure nothing is mutated and results are expected.


deploy/olm-catalog/watermarkpodautoscaler/manifests/watermarkpodautoscaler.clusterserviceversion.yaml
```diff
@@ -39,7 +39,7 @@ metadata:
         }
       ]
     capabilities: Basic Install
-  name: watermarkpodautoscaler.v0.1.0
+  name: watermarkpodautoscaler.v0.2.0-17-g36a0081-dirty
```
clamoriniere, line 42: I suppose this is not attended

deploy/olm-catalog/watermarkpodautoscaler/manifests/watermarkpodautoscaler.clusterserviceversion.yaml
```diff
@@ -242,4 +242,5 @@ spec:
   maturity: alpha
   provider:
     name: Datadog
-  version: 0.1.0
+  replaces: watermarkpodautoscaler.v0.1.0
+  version: 0.2.0-17-g36a0081-dirty
```
clamoriniere, line 246: same here

deploy/olm-catalog/watermarkpodautoscaler/manifests/watermarkpodautoscaler.clusterserviceversion.yaml
CharlyF: yeah, it was added when I generated the olm, I wanted to circle back with Vincent tomorrow to know more

pkg/apis/datadoghq/v1alpha1/watermarkpodautoscaler_default.go
```diff
@@ -5,10 +5,14 @@
...
-import "fmt"
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/api/resource"
+)
 
 const (
-	defaultTolerance                       = 0.1
+	defaultTolerance                       = 100
```
clamoriniere, line 15: what is the unit?

clamoriniere: could you add a comment about this value? what should be the min-max for a tolerance value

pkg/apis/datadoghq/v1alpha1/watermarkpodautoscaler_default.go
```diff
@@ -56,13 +59,14 @@ func IsDefaultWatermarkPodAutoscaler(wpa *WatermarkPodAutoscaler) bool {
 	if wpa.Spec.Algorithm == "" {
 		return false
 	}
-	if wpa.Spec.Tolerance == 0 {
+	// TODO omitempty for the next values, check what an empty resource yields.
```
clamoriniere, line 62: Should we remove this TODO?

pkg/controller/watermarkpodautoscaler/replica_calculator.go
```diff
@@ -177,25 +177,26 @@ func (c *ReplicaCalculator) GetResourceReplicas(logger logr.Logger, target *auto
 
 func getReplicaCount(logger logr.Logger, currentReplicas, currentReadyReplicas int32, wpa *v1alpha1.WatermarkPodAutoscaler, name string, adjustedUsage float64, lowMark, highMark *resource.Quantity) (replicaCount int32, utilization int64) {
 	utilizationQuantity := resource.NewMilliQuantity(int64(adjustedUsage), resource.DecimalSI)
-	adjustedHM := float64(highMark.MilliValue()) + wpa.Spec.Tolerance*float64(highMark.MilliValue())
-	adjustedLM := float64(lowMark.MilliValue()) - wpa.Spec.Tolerance*float64(lowMark.MilliValue())
+	adjustedHM := float64(highMark.MilliValue() + highMark.MilliValue()*wpa.Spec.Tolerance.MilliValue()/1000)
```
clamoriniere, line 180: should we check first that `wpa.Spec.Tolerance` is not `nil` ?

clamoriniere (commented):
few comments about OLM generation, otherwise looks good :)

pkg/apis/datadoghq/v1alpha1/watermarkpodautoscaler_default.go
CharlyF: it's to be used later on as milliquantity, so we can get 10%

pkg/controller/watermarkpodautoscaler/replica_calculator.go
CharlyF: we will be okay with the defaulting and as it's not a pointer `Tolerance resource.Quantity`
and in `IsDefaultWatermarkPodAutoscaler` we check if the value is == 0, in which cas we default it to 10%.
I added a test case to be explicit about the fact that not specifying the tolerance does not make the spec invalid.
It is however the role of the defaulting methods to ensure we set it to `defaultTolerance`

deploy/olm-catalog/watermarkpodautoscaler/manifests/watermarkpodautoscaler.clusterserviceversion.yaml
vboulineau: It's from `VERSION?=$(shell git describe --tags --dirty)` in Makefile. OLM should be generated on tags or we need to change Makefile the way we did it in operator/EDS to take `VERSION` from environment if set.

vboulineau (approved):

PR #71, opened 2020-12-31
Add logic to rely on leaderelection from controller manager

CharlyF:
### What does this PR do?

The WPA as is can't run in HA with a health probe, because of the blocking channel to know if the controller is the leader.
In other projects we have introduced a metric on top of relying on the leaderelection logic in the controller manager (not starting the Runnables if the controller is not the leader), then we rely on the metrics endpoint for health.
I decided to add a health check to be used as a health probe and add the ability to enable or not the leader election.

Note that if the leader loses the election, the pod will be rotated. I find this behaviour suboptimal because it does not take full advantage of the concept. However, because the upstream lib does not have logic process re-election, it is what is best.

Case 1: delete the CM lock until the follower wins the election. The leader will be terminated:
```
{"level":"info","ts":1609368911.9738276,"logger":"wpa_controller","msg":"Target deploy","Request.Namespace":"default","Request.Name":"example3-watermarkpodautoscaler","replicas":0}
I1230 22:55:26.878601       1 leaderelection.go:277] failed to renew lease default/watermarkpodautoscaler-lock: timed out waiting for the condition
{"level":"error","ts":1609368926.8786643,"logger":"cmd","msg":"Manager exited non-zero","error":"leader election lost","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\twatermarkpodautoscaler/vendor/github.com/go-logr/zapr/zapr.go:128\nmain.main\n\twatermarkpodautoscaler/cmd/manager/main.go:120\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:204"}
```
Case 2: update the record in the configmap directly (not that this would happen realistically, but who knows), leader is terminated similarly:
```
I1230 22:52:51.035952       1 leaderelection.go:277] failed to renew lease default/watermarkpodautoscaler-lock: timed out waiting for the condition
{"level":"error","ts":1609368771.0360084,"logger":"cmd","msg":"Manager exited non-zero","error":"leader election lost","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\twatermarkpodautoscaler/vendor/github.com/go-logr/zapr/zapr.go:128\nmain.main\n\twatermarkpodautoscaler/cmd/manager/main.go:120\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:204"}
```
Case 3: controller becomes unresponsive:
- The health probe will kill it
- The follower will take over once the deadline is passed

The issue with the metric is that if a leader were to lose the leadership but not die! It would still submit the metrics as leader true which would be confusing. 
I did not introduce a metric because given the above, there is no way to get a `OnTransition` signal, thus for a leader to lose the election, the pod must be rotated (unfortunately).

### Motivation

Improve the HA

### Additional Notes

### Describe your test plan

ensure you can set the LE and the health ports:
```
      containers:
      - args:
        - --health-port=1234
        - enable-leader-election=true
        command:
        - watermarkpodautoscaler
```
with a health check
```
        livenessProbe:
          failureThreshold: 3
          httpGet:
            path: /healthz
            port: 1234
            scheme: HTTP
          initialDelaySeconds: 15
          periodSeconds: 15
          successThreshold: 1
          timeoutSeconds: 5
```
avoid:
```
  Warning  Unhealthy  14s   kubelet            Liveness probe failed: Get http://10.124.26.11:1224/healthz: dial tcp 10.124.26.11:1224: connect: connection refused
```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -6,10 +6,11 @@
 package main
 
 import (
-	"context"
 	"flag"
 	"fmt"
 	"os"
+
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
@@ -19,7 +20,6 @@ import (
 	"github.com/DataDog/watermarkpodautoscaler/version"
 
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
-	"github.com/operator-framework/operator-sdk/pkg/leader"
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
 	"github.com/spf13/pflag"
 
@@ -32,11 +32,13 @@ import (
 
 // Change below variables to serve metrics on different host or port.
 var (
-	metricsHost       = "0.0.0.0"
+	host              = "0.0.0.0"
 	metricsPort int32 = 8383
+	healthPort  int32 = 9440
 )
 var log = logf.Log.WithName("cmd")
 var printVersionArg bool
+var enableLeaderElection bool
 
 func main() {
 	// Add the zap logger flag set to the CLI. The flag set must
@@ -47,7 +49,8 @@ func main() {
 	// controller-runtime)
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 	pflag.BoolVarP(&printVersionArg, "version", "v", printVersionArg, "print version")
-
+	pflag.BoolVar(&enableLeaderElection, "enable-leader-election", true, "Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
+	pflag.Int32Var(&healthPort, "health-port", healthPort, "Port to use for the health probe")
 	pflag.Parse()
 
 	// Use a zap logr.Logger implementation. If none of the zap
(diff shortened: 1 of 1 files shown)


cmd/manager/main.go
```diff
@@ -80,28 +83,21 @@ func main() {
...
-	}
-
 	// Create a new Cmd to provide shared dependencies and start components
 	mgr, err := manager.New(cfg, manager.Options{
-		Namespace:          namespace,
-		MapperProvider:     apiutil.NewDiscoveryRESTMapper,
-		MetricsBindAddress: fmt.Sprintf("%s:%d", metricsHost, metricsPort),
+		Namespace:              namespace,
+		MapperProvider:         apiutil.NewDiscoveryRESTMapper,
+		MetricsBindAddress:     fmt.Sprintf("%s:%d", host, metricsPort),
```
clamoriniere, line 90: maybe `net.JoinHostPort` can be used here. but I don't remember if the port is an `int32`

clamoriniere (approved):

clamoriniere (commented):
Could you check if the helm chart and the deployment manifest have the RBACs that allow the controller to Get/Watch/Update the ConfigMap used for the leader election.

cmd/manager/main.go
CharlyF: That's been working well in my testings, I got some inspiration from other upstream controllers

PR #72, opened 2021-01-04
Adding logic to support a 0 factor to limit scaling events in either direction

CharlyF:
### What does this PR do?

Adding support for never downscaling or upscaling by simply allowing 0 as a scaling factor limit.

### Motivation

Convenient for workloads like STS that we do not want to downscale automatically or workloads we don't want to scale up (nodegroup size, expensive etc).
Note that this is not the same as setting the max/min to whatever the current is, whatever the value and the current, if you set a scalinglimitfactor of 0, the WPA will not scale in this direction.
Though, metrics can still be used to see what the recommendation would be.

Example:

```
{"level":"info","ts":1609791650.3780568,"logger":"wpa_controller","msg":"Reconciling WatermarkPodAutoscaler","Request.Namespace":"default","Request.Name":"example3-watermarkpodautoscaler"}
{"level":"info","ts":1609791650.3818717,"logger":"wpa_controller","msg":"Target deploy","Request.Namespace":"default","Request.Name":"example3-watermarkpodautoscaler","replicas":3}
{"level":"info","ts":1609791650.382264,"logger":"wpa_controller","msg":"getReadyPodsCount","full podList length":6,"toleratedAsReadyPodCount":3,"incorrectly targeted pods":0}
{"level":"info","ts":1609791650.3901093,"logger":"wpa_controller","msg":"Metrics from the External Metrics Provider","Request.Namespace":"default","Request.Name":"example3-watermarkpodautoscaler","metrics":[78981]}
{"level":"info","ts":1609791650.390147,"logger":"wpa_controller","msg":"Value is above highMark","Request.Namespace":"default","Request.Name":"example3-watermarkpodautoscaler","usage":"78981m","replicaCount":10,"currentReadyReplicas":3,"tolerance (%):":1,"adjustedHM":25250,"adjustedUsage":78981}
{"level":"info","ts":1609791650.3902085,"logger":"wpa_controller","msg":"Proposing replicas","Request.Namespace":"default","Request.Name":"example3-watermarkpodautoscaler","proposedReplicas":10,"metricName":"sinus{map[foo:bar]}","reference":"Deployment/default/sinus"}
{"level":"info","ts":1609791650.3902335,"logger":"wpa_controller","msg":"Upscaling rate higher than limit set by 'ScaleUpLimitFactor', capping the maximum upscale to 'maximumAllowedReplicas'","Request.Namespace":"default","Request.Name":"example3-watermarkpodautoscaler","scaleUpLimitFactor":"0.0","wpaMaxReplicas":7,"maximumAllowedReplicas":3}
{"level":"info","ts":1609791650.3902442,"logger":"wpa_controller","msg":"Returning replicas, condition and reason","Request.Namespace":"default","Request.Name":"example3-watermarkpodautoscaler","replicas":3,"condition":"ScaleUpLimit","reason":"the desired replica count is increasing faster than the maximum scale rate"}
{"level":"info","ts":1609791650.3902557,"logger":"wpa_controller","msg":"Normalized Desired replicas","Request.Namespace":"default","Request.Name":"example3-watermarkpodautoscaler","desiredReplicas":3}
{"level":"info","ts":1609791650.39027,"logger":"wpa_controller","msg":"Will not scale: number of replicas has not changed","Request.Namespace":"default","Request.Name":"example3-watermarkpodautoscaler"}
```

### Describe your test plan

spin up a WPA and modify the scaleup factor to 0, verify that the workload is not scaled and metrics are OK (see the above logs).


pkg/apis/datadoghq/v1alpha1/watermarkpodautoscaler_types.go
```diff
@@ -58,12 +58,14 @@ type WatermarkPodAutoscalerSpec struct {
 	UpscaleForbiddenWindowSeconds int32 `json:"upscaleForbiddenWindowSeconds,omitempty"`
 
 	// Percentage of replicas that can be added in an upscale event.
-	// Parameter used to be a float, in order to support the transition seamlessly, we validate that it is ]0;100] in the code.
-	ScaleUpLimitFactor resource.Quantity `json:"scaleUpLimitFactor,omitempty"`
+	// Parameter used to be a float, in order to support the transition seamlessly, we validate that it is [0;100] in the code.
```
xornivore, line 61: Finding this comment a bit confusing, is the `used to be a float` a reference to prior implementation. Should we change this comment for clarity?

pkg/apis/datadoghq/v1alpha1/watermarkpodautoscaler_default.go
```diff
@@ -91,6 +91,9 @@ func CheckWPAValidity(wpa *WatermarkPodAutoscaler) error {
 	if wpa.Spec.Tolerance.MilliValue() > 1000 || wpa.Spec.Tolerance.MilliValue() < 0 {
 		return fmt.Errorf("tolerance should be set as a quantity between 0 and 1, currently set to : %v, which is %.0f%%", wpa.Spec.Tolerance.String(), float64(wpa.Spec.Tolerance.MilliValue())/10)
 	}
+	if wpa.Spec.ScaleUpLimitFactor == nil || wpa.Spec.ScaleDownLimitFactor == nil {
+		return fmt.Errorf("scaleuplimitfactor and scaledownlimitfactor can't be nil, make sure the WPA spec is defaulted")
```
xornivore, line 95: It might be worth rephrasing the second part about "defaulted", if a user sees this somewhere in the status it maybe misleading. Unless this error never surfaces that way?

xornivore (commented):
Looks great, some small comments to clarify.

Would it be possible to add e2e tests for WPA?

pkg/apis/datadoghq/v1alpha1/watermarkpodautoscaler_types.go
CharlyF: ah yeah I can lol

pkg/apis/datadoghq/v1alpha1/watermarkpodautoscaler_default.go
CharlyF: it should never been surfaced, because the WPA spec should be defaulted with the DefaultWPA method. This is more for the unit tests

PR #73, opened 2021-01-18
Make importable as a go module

albertvaka:
### What does this PR do?

Use a version that exists for `k8s.io/metrics` and `k8s.io/kube-aggregator`

Also ran `go mod tidy` which removed a dependency and a few go.sum entries.

### Motivation

Even though we have a `replace` directive for `k8s.io/metrics` and `k8s.io/kube-aggregator`, when importing a module the `replace` directives aren't taken into account so you would get an error if the version doesn't exist.


albertvaka:
I see that we import `k8s.io/kubernetes`, so that probably makes this not importable anyway due to the transitive dependencies there... closing this. Would be nice to review if we really must depend on `k8s.io/kubernetes` both here and in `datadog-agent`, since it's not meant to be imported and does cause problems with go modules.

clamoriniere:
hi @albertvaka 

@CharlyF is working on updating the project's dependencies to use operator-sdk v1. Thanks to this update we should be able to remove te dependency on `kubernetes/kubernetes`

PR #74, opened 2021-02-08
Migration to Operator 1.0.0

CharlyF:
### What does this PR do?

Updating the WPA codebase to use the Operator 1.0.0.
See the migration guidelines [here](https://sdk.operatorframework.io/docs/building-operators/golang/migration/).

Each commit should be scoped over a step of the migration.

- [x] New config folder
- [x] Core logic in `/controllers` and main.go moved to root
- [x] Tests are in the `/controllers` folder, e2e in `/controllers/test`. E2E are with github too, to align with the ExtendedDaemonset
- [x] Updated the version of go for gitlab runners because of module support

### Describe your test plan

- [ ] Clone the repository from scratch
- [ ] Run the `make` functions and make sure they work
- [ ] Run the E2E
- [ ] Run the integration tests and unit tests
- [ ] Build an image `make docker-build-ci` use it in a cluster that already has WPAs

### To Do

- [ ] We need to migrate all the unit tests to the new format (gomega and  ginkgo) to be consistent
- [x] Update the image name in the kustomization.yaml (Once we merge this, I am happy to update)
- [x] Update the license header to be 2021
- [ ] Add code coverage


api/v1alpha1/doc.go
```diff
@@ -1,7 +1,7 @@
 // Unless explicitly stated otherwise all files in this repository are licensed
 // under the Apache License Version 2.0.
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
-// Copyright 2016-2019 Datadog, Inc.
+// Copyright 2016-2020 Datadog, Inc.
```
clamoriniere, line 4:
```suggestion
// Copyright 2016-2021 Datadog, Inc.
```
;)

api/v1alpha1/groupversion_info.go
```diff
@@ -0,0 +1,25 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-2020 Datadog, Inc.
```
clamoriniere, line 4: I think you can do a big `sed`

```suggestion
// Copyright 2016-2021 Datadog, Inc.
```

bundle/manifests/watermarkpodautoscaler.clusterserviceversion.yaml
```diff
@@ -0,0 +1,189 @@
+apiVersion: operators.coreos.com/v1alpha1
```
clamoriniere, line 1: you can test the render with this link: https://operatorhub.io/preview

api/v1alpha1/doc.go
L3n41c: I was suggested in DataDog/datadog-agent#7332 to use `present` instead of `2021` in the copyright mentions.
It seems to be what we do for other repositories.
```suggestion
// Copyright 2016-present Datadog, Inc.
```

CharlyF: I guess I will  complete the To Do now heh

bundle/manifests/watermarkpodautoscaler.clusterserviceversion.yaml
CharlyF: I did some boilerplate for the operator but I think taking this to the finish line will require more work (I will add a card, as part of the todo next)

test/cluster-kind.yaml
```diff
@@ -0,0 +1,5 @@
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+name: wpa-e2e-test
+networking:
+  apiServerPort: 32001
```
clamoriniere, line 5: is it necessary?

clamoriniere (approved):
💯

test/cluster-kind.yaml
CharlyF: I think this is leftover from the previous e2e, not necessary I think.

go.mod
```diff
@@ -7,33 +7,34 @@ require (
...
-	k8s.io/kube-openapi v0.0.0-20200121204235-bf4fb3bd569c
+	k8s.io/kube-openapi v0.0.0-20200410145947-61e04a5be9a6
 	k8s.io/kubernetes v1.18.2
 	k8s.io/metrics v0.0.0
-	sigs.k8s.io/controller-runtime v0.6.0
+	sigs.k8s.io/controller-runtime v0.6.2
 )
 
 // Pinned to kubernetes-1.18.2
 replace (
```
vboulineau, line 30: As you removed the `operator-sdk` dependency, you should be able to remove the replace section entirely

pkg/client/clientset/versioned/clientset.go
```diff
@@ -1,7 +1,7 @@
 // Unless explicitly stated otherwise all files in this repository are licensed
```
vboulineau, line 1: The whole generated clientset `pkg/client` should be removed, it's not generated anymore.

test/e2e/utils/fakeapp.go
```diff
@@ -1,7 +1,7 @@
 // Unless explicitly stated otherwise all files in this repository are licensed
```
vboulineau, line 1: Idealy the `test` folder should be removed and its content moved to `controllers` or `pkg`

CharlyF: will move in a subsequent PR, as per the To Do section in the description of this PR.

vboulineau: Got it, for me unit tests are like `controllers/replica_calculator_test.go`, which we don't need to change. These are more e2e tests as they rely on complete system running.

pkg/client/clientset/versioned/clientset.go
CharlyF: great catch, I actually believed this was still used for the operator behind the scene.

go.mod
CharlyF: unfortunately no, I pull in `k8s.io/kubernetes`, so there are some indirect deps I need:
```
go: k8s.io/kubernetes@v1.18.2 requires
	k8s.io/sample-apiserver@v0.0.0: reading k8s.io/sample-apiserver/go.mod at revision v0.0.0: unknown revision v0.0.0
[...]
go: k8s.io/kubernetes@v1.18.2 requires
	k8s.io/legacy-cloud-providers@v0.0.0: reading k8s.io/legacy-cloud-providers/go.mod at revision v0.0.0: unknown revision v0.0.0
```

This is something I had identified however, and we have a card in the backlog to move away from k8s/k8s to only get the versioned client. That should help with the deps.

vboulineau (approved):

PR #75, opened 2021-02-11
Add metric_name label to the replicas_scaling_proposal

trazfr:
### What does this PR do?

When using several metrics as input, we now expose several Prometheus metrics `replicas_scaling_proposal`.

### Motivation

When using several metrics in input, `replicas_scaling_proposal` only reflects the last one defined in k8s' `spec.metrics`.

### DIFF
--- a/controllers/metrics.go
+++ b/controllers/metrics.go
@@ -128,6 +128,7 @@ var (
 			resourceNamespacePromLabel,
 			resourceNamePromLabel,
 			resourceKindPromLabel,
+			metricNamePromLabel,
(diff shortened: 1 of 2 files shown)


CharlyF:
Could confirm that it works well, that's a great contribution. Thanks @trazfr!
```
wpa_controller_replicas_scaling_proposal{metric_name="sinus",resource_kind="Deployment",resource_name="sinus",resource_namespace="default",wpa_name="example3-watermarkpodautoscaler"} 11
```

PR #76, opened 2021-03-08
Scale replicas by a configurable interval

zrosenberg:
### What does this PR do?

This change exposes a new configuration option, `ReplicaScaleInterval` which controls how many replicas will be added at a time. When calculating the desired replicas, we use the existing algorithm to determine the desired number of replicas, and then we round _up_ to the nearest `ReplicaScaleInterval`. 

I didn't touch the interaction with the maxReplicas at all, so if your `maxReplicas` value is not divisble by the `ReplicaScaleInterval`, the max will win. 

### Motivation

I want to use WPA with kafka, with a specific number of replicas at any given moment. This change would allow us to always have a number of replicas that is divisible by 3, 5, etc.

### Additional Notes

The churn in `api/v1alpha1/zz_generated.openapi.go` doesn't seem quite right to me, I ran `make generate` and that came out. 

I want this to have a default value of 1, so that no behavior changes with this change unless you explicitly set a new value. That's my understanding of what `api/v1alpha1/watermarkpodautoscaler_default.go` does, but I'm not 100% sure since I needed to specify a value in the tests. Maybe that's just how the tests are setup.

### Describe your test plan

Please let me know if I should add any additional tests!

### DIFF
--- a/api/v1alpha1/watermarkpodautoscaler_types.go
+++ b/api/v1alpha1/watermarkpodautoscaler_types.go
@@ -66,6 +66,10 @@ type WatermarkPodAutoscalerSpec struct {
 	// ScaleDownLimitFactor == 0 means that downscaling will not be allowed for the target.
 	ScaleDownLimitFactor *resource.Quantity `json:"scaleDownLimitFactor,omitempty"`
 
+	// Number of replicas to scale by at a time. When set, replicas added or removed must be a multiple of this parameter.
+	// Allows for special scaling patterns, for instance when an application requires a certain number of pods in multiple
+	ReplicaScalingInterval *int32 `json:"replicaScalingInterval,omitempty"`
+
 	// Parameter used to be a float, in order to support the transition seamlessly, we validate that it is ]0;1[ in the code.
 	Tolerance resource.Quantity `json:"tolerance,omitempty"`
 
--- a/api/v1alpha1/watermarkpodautoscaler_types.go
+++ b/api/v1alpha1/watermarkpodautoscaler_types.go
@@ -68,6 +68,7 @@ type WatermarkPodAutoscalerSpec struct {
 
 	// Number of replicas to scale by at a time. When set, replicas added or removed must be a multiple of this parameter.
 	// Allows for special scaling patterns, for instance when an application requires a certain number of pods in multiple
+	// +kubebuilder:validation:Minimum=1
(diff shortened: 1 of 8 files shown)


api/v1alpha1/watermarkpodautoscaler_types.go
```diff
@@ -66,6 +66,10 @@ type WatermarkPodAutoscalerSpec struct {
 	// ScaleDownLimitFactor == 0 means that downscaling will not be allowed for the target.
 	ScaleDownLimitFactor *resource.Quantity `json:"scaleDownLimitFactor,omitempty"`
 
+	// Number of replicas to scale by at a time. When set, replicas added or removed must be a multiple of this parameter.
+	// Allows for special scaling patterns, for instance when an application requires a certain number of pods in multiple
+	ReplicaScalingInterval *int32 `json:"replicaScalingInterval,omitempty"`
```
clamoriniere, line 71: Could you add the followig `kubebuilder` comment.

```suggestion
         // +kubebuilder:validation:Minimum=1
	ReplicaScalingInterval *int32 `json:"replicaScalingInterval,omitempty"`
```

Also `ScaleReplicaFactor` can be a better name. 
@CharlyF WDYT?

clamoriniere (commented):
This feature make sens. 
I have added few comments. I think it can be good to add more unit-tests to have at least:
* scale up no triggered
* scale up triggered with the ReplicasScalingInterval applied
* scale down triggered with the ReplicasScalingInterval applied
* scale down not triggered.
and some corner cases:
* scale down triggered but the result will be negative. 
* scale down triggered that will be under the min replica.
* scale up triggered that will be over the max replica.

I have also a more general question:
* does the ReplicaScalingInterval should be apply for scale up and scale down? Do you see some use cases where the value require to be different?

zrosenberg:
@clamoriniere I've added a couple tests, but for your corner cases I don't see a way to add those tests that would actually exercise my code, since its all contained in the Replica Calculator and the `watermarkpodautoscaler_controller_test.go` seems like it stubs the results of the calculator everywhere. The current tests are pretty tightly unit tested per class.

With this change, the behavior for crossing either maxReplicas or minReplicas either way is preserved. This changes the way that the desired replicas is calculated, but it will have the same effect if it exceeds the limits. Hope that makes sense

zrosenberg:
@clamoriniere @CharlyF Where are we with this?

clamoriniere (approved):
Thank you both for this contribution 👍

PR #87, opened 2021-05-11
Fix WPA dry-run status in kubectl output

clamoriniere:
### What does this PR do?

Update the ` +kubebuilder:printcolumn` on top of `type WatermarkPodAutoscaler struct `

```golang
// +kubebuilder:printcolumn:name="dry-run",type="string",JSONPath=".status.conditions[?(@.type==\"DryRun\")].status"
// ...
type WatermarkPodAutoscaler struct 
```

### Motivation

Fix WPA dry-run status in kubectl output.

### Additional Notes

### Describe your test plan

Create a WPA without the dry-run mode. the `dry-run` status with `kubectl get wpa` should be equal to `false`


CharlyF (approved):
good catch

PR #88, opened 2021-05-12
Fix rbacs and make deploy

ahmed-mez:
### What does this PR do?

- Fix rbac's api groups
- Fix role binding name

### Motivation

Fix `make deploy`

### Describe your test plan

`make deploy` should succeed + the controller works without any rbac errors

### DIFF
--- a/config/rbac/role_binding.yaml
+++ b/config/rbac/role_binding.yaml
@@ -5,7 +5,7 @@ metadata:
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: watermarkpodautoscaler-controller-manager
+  name: watermarkpodautoscaler-manager
 subjects:
 - kind: ServiceAccount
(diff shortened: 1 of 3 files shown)


config/rbac/role_binding.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: watermarkpodautoscaler-controller-manager
+  name: watermarkpodautoscaler-manager
```
clamoriniere, line 8: is it a manual change you did?

ahmed-mez: yes, should it be done somewhere else?

clamoriniere: I think for this one is manual, but it can be referenced by some kustomize logic. I checked and I did found anything.
So all good 👍

clamoriniere (approved):

PR #90, opened 2021-05-24
Fix bug pertaining to timestamp override

CharlyF:
### What does this PR do?

The core change here is removing  `now = metricTimestamp`

The timestamp of the datapoint of a metric **should not** be used as reference to compute whether we are allowed to scale.
What will happen is if the metric is 90s old (which is within range of acceptability) comparing the LastScalingTime to it and to the forbidden windows gives an incorrect result.
See the example `Forbidden window uses the right timestamp` in the tests.

Also fixed the tests, without the Reactor (get specifically) we would get empty objects, so everything would pass because it would be comparing nil objects.

### Motivation

Fix bug.

### Additional Notes

### Describe your test plan

### DIFF
--- a/controllers/watermarkpodautoscaler_controller_test.go
+++ b/controllers/watermarkpodautoscaler_controller_test.go
@@ -21,20 +21,18 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	autoscalingv1 "k8s.io/api/autoscaling/v1"
 	"k8s.io/api/autoscaling/v2beta1"
-	"k8s.io/api/autoscaling/v2beta2"
 	corev1 "k8s.io/api/core/v1"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/api/meta/testrestmapper"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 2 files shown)


controllers/watermarkpodautoscaler_controller_test.go
```diff
@@ -307,37 +304,49 @@ func newRequest(ns, name string) reconcile.Request {
...
-	return nil, time.Time{}, nil
-}
-
-//
-//// GetObjectMetric gets the given metric (and an associated timestamp) for the given
-//// object in the given namespace
-func (f fakeMetricsClient) GetObjectMetric(metricName string, namespace string, objectRef *v2beta2.CrossVersionObjectReference, metricSelector labels.Selector) (int64, time.Time, error) {
-	return 0, time.Time{}, nil
+func addUpdateReactor(s *fakescale.FakeScaleClient) {
+	s.AddReactor("update", "deployments", func(rawAction testcore.Action) (handled bool, ret runtime.Object, err error) {
```
vboulineau, line 308: For consistency between `addUpdateReactor` and `addGetReactor`, either put all reactor functions inline (update) or split (get). I have no strong preference between the two.

vboulineau (approved):

PR #112, opened 2021-08-16
Add latest condition and latest scale to `kubectl get wpa` output

davidor:
### What does this PR do?

Adds new columns to the `kubectl get wpa` output:
- `condition` ("ScalingActive", "AbleToScale", "ScalingLimited")
- `condition status` (true or false)
- `last scale` (time passed since last scale).

To be able to show the latest condition, I needed to keep them sorted. That's why `kubectl get wpa -o yaml` will now always show them sorted by "LatestTransitionTime" (most recent first).

### Describe your test plan

Deploy an WPA. Check that the 3 new columns mentioned above appear in `kubectl get wpa` and show correct information. For `condition` and `condition status`, you can check that the one shown corresponds to the most recent condition shown in `kubectl get wpa -o yaml`.


CharlyF (approved):
great work.

PR #113, opened 2021-09-01
Adds labels from individual WPAs to the rescale log message.

zrosenberg:
### What does this PR do?

Adds all labels on the WPA resource as log attributes, alongside whatever other attributes are already present.

### Motivation

What inspired you to submit this pull request?

I'm interested in being able to use the logs to metrics pipeline for the successful rescale log message to generate useful metrics, and have them be sliceable by a couple different dimensions. I realized that all the useful dimensions I want are already expressed as labels, so that the label join gets those same dimensions on wpa metrics.

### Additional Notes

### Describe your test plan


zrosenberg:
cc @clamoriniere @CharlyF

api/v1alpha1/watermarkpodautoscaler_types.go
```diff
@@ -204,3 +204,17 @@ type WatermarkPodAutoscalerList struct {
 func init() {
 	SchemeBuilder.Register(&WatermarkPodAutoscaler{}, &WatermarkPodAutoscalerList{})
 }
+
+// GetLogAttrs takes a list of log attrs that would be sent to logger.info, and adds all key value label pairs from
+// the WPA spec. logAttrs should be pairs of key/values, with each key being a string.
+func (wpa *WatermarkPodAutoscaler) GetLogAttrs(logAttrs ...interface{}) []interface{} {
+	var logAttributes []interface{}
+	for k,v := range wpa.ObjectMeta.Labels {
```
CharlyF, line 212: I think we should have a whitelist of labels that we want to add, that could be configurable

controllers/watermarkpodautoscaler_controller.go
```diff
@@ -364,7 +364,7 @@ func canScale(logger logr.Logger, backoffUp, backoffDown bool, currentReplicas,
 		logger.Info("Will not scale: number of replicas has not changed")
 		return false
 	}
-	logger.Info("Cooldown status", "backoffUp", backoffUp, "backoffDown", backoffDown, "desiredReplicas", desiredReplicas, "currentReplicas", currentReplicas)
+	logger.Info("Cooldown status","backoffUp", backoffUp, "backoffDown", backoffDown, "desiredReplicas", desiredReplicas, "currentReplicas", currentReplicas)
```
CharlyF, line 367: not here?

api/v1alpha1/watermarkpodautoscaler_types.go
```diff
@@ -204,3 +205,30 @@ type WatermarkPodAutoscalerList struct {
...
+// logAttrs should be pairs of key/values, with each key being a string.
+// This func returns a slice! To pass this as variadic parameters to logger.info, you need the ellipses, so use wpa.GetLogAttrs()...
+func (wpa *WatermarkPodAutoscaler) GetLogAttrs(logAttrs ...interface{}) []interface{} {
+	var logAttributes []interface{}
+
+	for _, v := range logAttrs {
+		logAttributes = append(logAttributes, v)
+	}
+
+	customAttrsStr := wpa.ObjectMeta.Annotations["ad.datadoghq.com/attributes"]
```
clamoriniere, line 220: I would prefer to use another key for the annotation because `ad.datadoghq.com` is managed by the agent and it apply for autodiscovery features, which is not the case here. 
So I think you could use `wpa.datadoghq.com/logs-attributes` instead.

clamoriniere: also could you define a `const string` to store this annotation key.

api/v1alpha1/watermarkpodautoscaler_types.go
```diff
@@ -204,3 +205,30 @@ type WatermarkPodAutoscalerList struct {
 func init() {
 	SchemeBuilder.Register(&WatermarkPodAutoscaler{}, &WatermarkPodAutoscalerList{})
 }
+
+// GetLogAttrs takes a list of log attrs that would be sent to logger.info, and adds user specified key value attribute
+// pairs from the ad.datadoghq.com/attributes annotation of the WPA spec.
+// logAttrs should be pairs of key/values, with each key being a string.
+// This func returns a slice! To pass this as variadic parameters to logger.info, you need the ellipses, so use wpa.GetLogAttrs()...
+func (wpa *WatermarkPodAutoscaler) GetLogAttrs(logAttrs ...interface{}) []interface{} {
```
clamoriniere, line 213: I would prefer to have a function in the `controllers` package like:

```golang
func GetLogAttrsFromWPA(wpa *apiv1alpha1.WatermarkPodAutoscaler) []interface{} {
  // ...
}
```

because the `v1alpha1` package is used to generate CRD definition.
So I would prefer that we keep it only with resource definition

controllers/watermarkpodautoscaler_controller.go
```diff
@@ -252,7 +252,7 @@ func (r *WatermarkPodAutoscalerReconciler) reconcileWPA(ctx context.Context, log
 	if rescale {
 		setCondition(wpa, autoscalingv2.AbleToScale, corev1.ConditionTrue, datadoghqv1alpha1.ConditionReasonReadyForScale, "the last scaling time was sufficiently old as to warrant a new scale")
 		if wpa.Spec.DryRun {
-			logger.Info("DryRun mode: scaling change was inhibited", "currentReplicas", currentReplicas, "desiredReplicas", desiredReplicas)
+			logger.Info("DryRun mode: scaling change was inhibited", wpa.GetLogAttrs("currentReplicas", currentReplicas, "desiredReplicas", desiredReplicas))
```
clamoriniere, line 255: instead of updating every `logger.Info()` call.
you can add at the beginning of the function `Reconcile()` [here](https://github.com/DataDog/watermarkpodautoscaler/pull/113/files#diff-70d5250d0b22ca57247df02274172975492cd9862cc800a40ec91644a462d9ddR109)

the logs attributes
```golang
log = log.WithValues(GetLogAttrsFromWPA(instance))
```
this values will be added to any log lines.

clamoriniere (commented):
I added few comments.

zrosenberg:
@clamoriniere Updated! thanks

controllers/test/get_wpa_log_attrs_test.go
```diff
@@ -0,0 +1,40 @@
...
+	logAttrs := controllers.GetLogAttrsFromWpa(wpa)
+
+	validateVariadicExpansion(t, []interface{}{}, logAttrs...)
+}
+
+func validateVariadicExpansion(t *testing.T, expected []interface{}, actual ...interface{}) {
+	for i, v := range expected {
+		assert.Equal(t, v, actual[i])
+	}
+}
```
clamoriniere, line 40:
```suggestion
}

```

controllers/watermarkpodautoscaler_controller.go
```diff
@@ -52,6 +53,7 @@ import (
 
 const (
 	defaultSyncPeriod = 15 * time.Second
+	logAttributesAnnotation = "wpa.datadoghq.com/logs-attributes"
```
clamoriniere, line 56:
```suggestion
	logAttributesAnnotationKey = "wpa.datadoghq.com/logs-attributes"
```

controllers/watermarkpodautoscaler_controller.go
```diff
@@ -355,11 +358,11 @@ func shouldScale(logger logr.Logger, wpa *datadoghqv1alpha1.WatermarkPodAutoscal
 		transitionCountdown.With(prometheus.Labels{wpaNamePromLabel: wpa.Name, transitionPromLabel: "upscale", resourceNamespacePromLabel: wpa.Namespace, resourceNamePromLabel: wpa.Spec.ScaleTargetRef.Name, resourceKindPromLabel: wpa.Spec.ScaleTargetRef.Kind}).Set(0)
 	}
 
-	return canScale(logger, backoffUp, backoffDown, currentReplicas, desiredReplicas)
+	return canScale(logger, backoffUp, backoffDown, currentReplicas, desiredReplicas, wpa)
```
clamoriniere, line 361: since you added the logs attributes in the `reconcile()` function.
you don't need anymore to provide the `wpa` to this function

clamoriniere (commented):
small additional nit comments

zrosenberg:
@clamoriniere Updated, and green. 

When this is ready to merge, will it be possible to get it into production quickly?

clamoriniere (approved):

PR #115, opened 2021-09-13
Fix CRD definition and validation

clamoriniere:
### What does this PR do?

* Fix CRD definition to pass the open-api validation
* Update CRD manifest in the helm chart.
* fix kubebuilder download link

### Motivation
When we tried to deploy the current CRD from the main branch with `make deploy` the following was returned by the api server
```
Error: CustomResourceDefinition.apiextensions.k8s.io "watermarkpodautoscalers.datadoghq.com" is invalid: [spec.validation.openAPIV3Schema.properties[spec].properties[metrics].items.x-kubernetes-map-type: Invalid value: "null": must be atomic as item of a list with x-kubernetes-list-type=set, spec.validation.openAPIV3Schema.properties[status].properties[conditions].items.x-kubernetes-map-type: Invalid value: "null": must be atomic as item of a list with x-kubernetes-list-type=set, spec.validation.openAPIV3Schema.properties[status].properties[currentMetrics].items.x-kubernetes-map-type: Invalid value: "null": must be atomic as item of a list with x-kubernetes-list-type=set]
```

### Additional Notes

### Describe your test plan

Try to install the WPA controller with the helm chart in a recent cluster `k8s 1.21` for example.
The install should work.

### DIFF
--- a/api/v1alpha1/watermarkpodautoscaler_types.go
+++ b/api/v1alpha1/watermarkpodautoscaler_types.go
@@ -88,6 +88,7 @@ type WatermarkPodAutoscalerSpec struct {
 	// and will set the desired number of pods by using its Scale subresource.
 	ScaleTargetRef CrossVersionObjectReference `json:"scaleTargetRef"`
 	// specifications that will be used to calculate the desired replica count
+	// +optional
 	// +listType=set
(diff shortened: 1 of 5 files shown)


CharlyF (approved):
LGTM!

PR #121, opened 2021-11-18
Prevent downscale when value over HWM if some pods are unready

CharlyF:
### What does this PR do?

With the current logic, if the metric is above the high watermark but you have some unready replicas, you can end up with fewer replicas than originally.
This is a problem for applications that take time to be ready for instance as it allows effectively downscaling when the scenario ought to yield a scale up event.

The expectation here is that even if some pods never become ready, we will keep at least as many pods as the status of the target has and this until the recommendation becomes higher than the current number of ready replicas.
 
I also fixed some tests that reflected this (incorrect, but designed) behavior. 
### Motivation

Better support for edge cases

### Additional Notes

if you have 5 ready replicas out of 10 in the spec of your deployment, 5 is used to compute the recommendation (based on the ratio of value/High Watermark). If the ratio is <2, this would trigger a scaling event to a value under 10. This is obviously undesired as it at best removes unready replicas but at worse remove ready ones.

### Describe your test plan

Create a deployment with a few unready replicas (could work with taints) and configure the WPA to trigger a scale up event, make sure you do not end up with fewer replicas than originally and see the log:
```
"Recommendation is lower than current number of replicas while attempting to upscale, aborting", "replicaCount", replicaCount, "currentReadyReplicas", currentReadyReplicas
```


clamoriniere (approved):
🚀

PR #127, opened 2021-12-29
Introduce option to specify minimum ratio of available pods

CharlyF:
### What does this PR do?

Adds an option to specify the minimum percentage of pods from a target that need to be available in order for the controller to scale a target.
Note that the amount of ready replicas does not take into account replicas that are not properly labled (e.g. we only consider the replicas from the target in the WPA spec). We also consider replicas in pending during `ReadinessDelaySeconds` s.

### Motivation

Better control of edge cases. Works as a fast fail.

### Additional Notes

### Describe your test plan

Scale rapidly a deployment that takes a long time to be marked available (use the probes) with a high ReadinessDelaySeconds to make sure that the controller stops if there are too many pending replicas.

### DIFF
--- a/api/v1alpha1/watermarkpodautoscaler_types.go
+++ b/api/v1alpha1/watermarkpodautoscaler_types.go
@@ -93,6 +93,9 @@ type WatermarkPodAutoscalerSpec struct {
 	Metrics []MetricSpec `json:"metrics,omitempty"`
 	// +kubebuilder:validation:Minimum=1
 	MinReplicas *int32 `json:"minReplicas,omitempty"`
+	// MinAvailableReplicaPerc indicates the minimum percentage of replicas that need to be available in order for the
+	// controller to autoscale the target.
+	MinAvailableReplicaPerc int32 `json:"minAvailableReplicaPerc,omitempty"`
 	// +kubebuilder:validation:Minimum=1
 	MaxReplicas int32 `json:"maxReplicas,omitempty"`
 	// +kubebuilder:validation:Minimum=1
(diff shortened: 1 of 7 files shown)


config/crd/bases/v1beta1/datadoghq.com_watermarkpodautoscalers.yaml
```diff
@@ -305,6 +311,7 @@ spec:
               minimum: 1
               type: integer
           required:
+          - minAvailableReplicaPerc
```
L3n41c, line 314: Is making a new field mandatory considered as a breaking change ?
(I.e.: currently existing WPA don’t have a `minAvailableReplicaPerc` setting)

L3n41c: Is there a default value for `minAvailableReplicaPerc` if it’s not set ?
If it’s 0, there’s a risk of blocking all WPA as soon as a single pod land on an unhealthy node.

api/v1alpha1/watermarkpodautoscaler_types.go
```diff
@@ -93,6 +93,9 @@ type WatermarkPodAutoscalerSpec struct {
 	Metrics []MetricSpec `json:"metrics,omitempty"`
 	// +kubebuilder:validation:Minimum=1
 	MinReplicas *int32 `json:"minReplicas,omitempty"`
+	// MinAvailableReplicaPerc indicates the minimum percentage of replicas that need to be available in order for the
+	// controller to autoscale the target.
+	MinAvailableReplicaPerc int32 `json:"minAvailableReplicaPerc,omitempty"`
```
clamoriniere, line 98: should we put the full name 
```suggestion
	MinAvailableReplicaPercentage int32 `json:"minAvailableReplicaPercentage,omitempty"`
```

clamoriniere: also maybe add kubebuilder comment constraint to limit the allow value

clamoriniere (approved):

PR #129, opened 2022-01-10
[controller] Increase requeue delay for "scale not found" errors

davidor:
### What does this PR do?

Increases the requeue delay for "scale not found" errors.

### Describe your test plan

### DIFF
--- a/controllers/watermarkpodautoscaler_controller.go
+++ b/controllers/watermarkpodautoscaler_controller.go
@@ -55,6 +55,10 @@ import (
 const (
 	defaultSyncPeriod          = 15 * time.Second
 	logAttributesAnnotationKey = "wpa.datadoghq.com/logs-attributes"
+	scaleNotFoundErr           = "scale not found"
+
+	defaultRequeueDelay       = time.Second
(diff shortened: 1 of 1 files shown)


controllers/watermarkpodautoscaler_controller.go
```diff
@@ -302,6 +303,16 @@ func (r *WatermarkPodAutoscalerReconciler) reconcileWPA(ctx context.Context, log
 	return r.updateStatusIfNeeded(ctx, wpaStatusOriginal, wpa)
 }
 
+func requeueAfterForWPAErrors(err error) time.Duration {
+	// We don't expect this error to be recovered after 1s, so define a longer
+	// delay.
+	if strings.Contains(err.Error(), scaleNotFoundErr) {
+		return 10 * time.Second
```
CharlyF, line 310: let's specify these two return values as const (here and l:313).

CharlyF (changes requested):

controllers/watermarkpodautoscaler_controller.go
davidor: Changed :+1:

CharlyF (approved):

PR #130, opened 2022-01-24
Updating helm chart

CharlyF:
### What does this PR do?

Upon merging releasing 0.4.0, we update the chart.

### Motivation
seamless upgrade experience for users.

### Additional Notes

### Describe your test plan

### DIFF
--- a/chart/watermarkpodautoscaler/templates/datadoghq.com_watermarkpodautoscalers_crd.yaml
+++ b/chart/watermarkpodautoscaler/templates/datadoghq.com_watermarkpodautoscalers_crd.yaml
@@ -7,6 +7,12 @@ metadata:
   name: watermarkpodautoscalers.datadoghq.com
 spec:
   additionalPrinterColumns:
+  - JSONPath: .status.conditions[0].type
+    name: condition
+    type: string
+  - JSONPath: .status.conditions[0].status
+    name: condition status
+    type: string
   - JSONPath: .status.currentMetrics[*].external.currentValue..
     name: value
     type: string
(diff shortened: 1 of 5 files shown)


CharlyF:
Merge when 0.4.o is released.

chart/watermarkpodautoscaler/templates/datadoghq.com_watermarkpodautoscalers_crd.yaml
```diff
@@ -25,9 +31,12 @@ spec:
   - JSONPath: .spec.maxReplicas
     name: max replicas
     type: integer
-  - JSONPath: .spec.dryRun
+  - JSONPath: .status.conditions[?(@.type=="DryRun")].status
```
clamoriniere, line 34: This is what I was talking about during the standup.

chart/watermarkpodautoscaler/templates/datadoghq.com_watermarkpodautoscalers_crd.yaml
```diff
@@ -7,6 +7,12 @@ metadata:
   name: watermarkpodautoscalers.datadoghq.com
 spec:
   additionalPrinterColumns:
+  - JSONPath: .status.conditions[0].type
+    name: condition
+    type: string
+  - JSONPath: .status.conditions[0].status
+    name: condition status
+    type: string
```
clamoriniere, line 15: how can we be sure it is always the first item in the conditions slice?

chart/watermarkpodautoscaler/values.yaml
```diff
@@ -29,6 +29,11 @@ labelsAsTags: []
 podSecurityContext: {}
 # fsGroup: 2000
 
+ddProfiling:
+  enabled: false
+
+workers: 1
```
clamoriniere, line 35: IMO to be backward compatible the `workers` value should be empty. because here you will always add the new args that doesn't exist on previous version.

OR, you add an `if` in the template file. Something like

```yaml
            {{- if no eq .Values.workers 1 }}
             - -workers={{ .Values.workers }}
            {{- end }}     
```

PR #138, opened 2022-02-16
fix: improve how we filter pods from other deployments

clamoriniere:
### What does this PR do?

previously the way we computed the list of pod ready corresponding
to a WPA was not enough precise. In some case if 2 deployments in
the same Namespace was using the same Selector can trigger a bug
in the wpa where both deployment's pods are taken into account when
calculating the current replicas value.

### Motivation

What inspired you to submit this pull request?

### Additional Notes

### Describe your test plan

### DIFF
--- a/controllers/replica_calculator.go
+++ b/controllers/replica_calculator.go
@@ -47,7 +47,6 @@ type ReplicaCalculator struct {
 
 // NewReplicaCalculator returns a ReplicaCalculator object reference
 func NewReplicaCalculator(metricsClient metricsclient.MetricsClient, podLister corelisters.PodLister) *ReplicaCalculator {
-
 	return &ReplicaCalculator{
 		metricsClient: metricsClient,
 		podLister:     podLister,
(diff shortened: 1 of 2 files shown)


controllers/replica_calculator.go
```diff
@@ -75,6 +74,12 @@ func (c *ReplicaCalculator) GetExternalMetricReplicas(logger logr.Logger, target
 		return ReplicaCalculation{}, fmt.Errorf("unable to get the number of ready pods across all namespaces for %v: %s", lbl, err.Error())
 	}
 
+	if currentReadyReplicas > target.Status.Replicas {
+		// if we enter in that condition it means that several Deployment use the same label selector in the same namespace.
+		logger.Info("The number of current ready Pods is higher that the current status.replicas", "currentReadyReplicas", currentReadyReplicas, "targetStatusReplicas", target.Status.Replicas)
```
L3n41c, line 79: Maybe the log message itself should contain the information that the most probable cause is a bad label selector, which is a user misconfiguration.
Written as it is, the log isn’t explicit about the fact that there is something to be fixed.
Also, the log misses the needed information to identify the guilty WPA: its name and namespace.

L3n41c (approved):

controllers/replica_calculator.go
clamoriniere: the `logger` inherit automatically from the ns/name of the current WPA, not need to add to it on each `logger.Info()` call that the beauty of thy logger :) 

something like
```golang
logger = logger.WithValue("wpa", wpa.Namespace + "/" + wpa.Name)
```
is made in the main `Reconcile()` function

clamoriniere: For the message I agree, I can be more explicit.

PR #145, opened 2022-04-28
[Draft] Add support for limiting downscale/upscale until metrics has evaluated below/above watermark for duration

richardartoul:
### What does this PR do?

This P.R adds support for additional configuration options that allow downscales/upscales to be constrained until the metric has evaluated below (or above) the corresponding watermark for a configurable duration. This will be useful for slowing down downscales and preventing downscales from occurring immediately as soon as a value drops below the low watermark which will improve reliability by preventing downscales during dependency failures and minor stalls.

### Motivation

We've had several issues where one of our dependencies temporarily fails and this causes the CPU of our service to dip which results in downscales which makes the subsequent recovery more difficult when the dependency is restored.

### DIFF
--- a/api/v1alpha1/watermarkpodautoscaler_types.go
+++ b/api/v1alpha1/watermarkpodautoscaler_types.go
@@ -193,10 +193,12 @@ type MetricSpec struct {
 // WatermarkPodAutoscalerStatus defines the observed state of WatermarkPodAutoscaler
 // +k8s:openapi-gen=true
 type WatermarkPodAutoscalerStatus struct {
-	ObservedGeneration *int64       `json:"observedGeneration,omitempty"`
-	LastScaleTime      *metav1.Time `json:"lastScaleTime,omitempty"`
-	CurrentReplicas    int32        `json:"currentReplicas"`
-	DesiredReplicas    int32        `json:"desiredReplicas"`
+	ObservedGeneration         *int64       `json:"observedGeneration,omitempty"`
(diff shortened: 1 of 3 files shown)


tgermain:
I've been thinking about that PR, what it's trying to achieve: make sure the input metrics of the wpa is above/below the high/low watermark for a long enough, configurable, duration before acting on it.

I'm still puzzled if it's the right place to do it or if we could do that on the metrics query side to have a better signal to drive the WPA.

See that [notebook](https://ddstaging.datadoghq.com/notebook/2429756/timothee-may-06-2022-10-32) where, using the same input metrics, but different option in `moving_rollup()`, I try to have a safer metric to feed to the WPA.

I'm not sure how feasable it is to add a `moving_rollup()` to all the metrics we use for the WPA, but that looks like another good proposal to solve that issue.

richardartoul:
@tgermain This problem cannot be solved with metrics queries or moving rollups, we've already tried. We *really* need this to be baked into the WPA directly.

The issue with putting it in the metric itself is that it constrains upscales. The way we want our WPAs to work is:

1. Upscale aggressively
2. Downscale very slowly

The current configuration simply does not make this possible. If we wanted to do this with metrics queries alone, we would need the WPA to support two different metrics, one for upscaling, and one for downscaling. That leads to a lot of extra complexity though in terms of reasoning about how those things will interact.

After several long discussions with Charly and our team I dont really see a better solution to this problem, and I think we need to get something merged urgently considering the number of incidents that are directly attributable to this problem.

PR #147, opened 2022-05-04
Don't upscale when downscaling with unready pods

khewonc:
### What does this PR do?

When a target has unready pods and needs to downscale, the WPA can propose a value greater than the number of ready pods and cause an upscale event. This change prevents the upscale event from happening by not scaling at all until the proposed value is smaller than the number of ready pods.

### Motivation

Edge case

### Describe your test plan

Create a WPA that scales on a deployment with several ready and unready pods. [Taints and tolerations](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/), [affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity), exaggerated resource requirements, etc. can be used to force unready pods. It's also easier if the WPA uses a smaller `scaleDownLimitFactor`, e.g. `15`, so that the WPA can't scale down many replicas at once.

Force a downscale using the WPA so that the WPA's proposed replicas value is larger than the number of ready replicas. For example, if we have 7 app pods, 5 ready and 2 unready, and we reduce the metric query value to much lower than the `lowWaterMark` value, the WPA will propose a replica value of 6 ([because of the `scaleDownLimitFactor`](https://github.com/DataDog/watermarkpodautoscaler/blob/799f4d9eeba8d64f9aac5fa52286b6d5757ea766/controllers/watermarkpodautoscaler_controller.go#L704)), which is larger than the 5 ready replicas. There should be no up/downscale events and this log should be present:

```
Maximum downscale recommendation is higher than current number of ready replicas, aborting
```

### DIFF
--- a/controllers/replica_calculator.go
+++ b/controllers/replica_calculator.go
@@ -27,9 +27,10 @@ import (
 
 // ReplicaCalculation is used to compute the scaling recommendation.
 type ReplicaCalculation struct {
-	replicaCount int32
-	utilization  int64
-	timestamp    time.Time
+	replicaCount  int32
+	utilization   int64
+	timestamp     time.Time
+	readyReplicas int32
 }
 
 // ReplicaCalculatorItf interface for ReplicaCalculator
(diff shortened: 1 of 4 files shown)


CharlyF (approved):
looks good - is there no way to go around adding `readyReplicas` in the `ReplicaCalculation` ?

CharlyF (approved):
looks good - is there no way to go around adding `readyReplicas` in the `ReplicaCalculation` ?

khewonc:
I could pass the `readyReplicas` around to different functions separately from the `ReplicaCalculation` if that would be better, or I could try separating out [this bit to grab the podlist](https://github.com/DataDog/watermarkpodautoscaler/blob/799f4d9eeba8d64f9aac5fa52286b6d5757ea766/controllers/replica_calculator.go#L64) and make an additional call for the podlist per reconcile. I haven't been able to think of alternatives beyond those so far 🤔

khewonc:
I could pass the `readyReplicas` around to different functions separately from the `ReplicaCalculation` if that would be better, or I could try separating out [this bit to grab the podlist](https://github.com/DataDog/watermarkpodautoscaler/blob/799f4d9eeba8d64f9aac5fa52286b6d5757ea766/controllers/replica_calculator.go#L64) and make an additional call for the podlist per reconcile. I haven't been able to think of alternatives beyond those so far 🤔

PR #150, opened 2022-05-24
Upgrade Go, k8s.io/*, openapi definitions

juliogreff:
### What does this PR do?

This is an attempt to solve a tricky dependency version puzzle in
datadog-agent. Since that repo depends on this one (and datadog-operator,
and extendeddaemonset) for type definitions, they need to be using
compatible openapi definitions. To get them, we also need to update
k8s.io to a recent enough version (but not too recent! dependency
issues). To be able to update k8s.io, we also need to upgrade to Go
1.17, since that's now required by controller-runtime.

### Describe your test plan

This only upgrades dependencies, so no change in behavior is expected. Check that main functionality of the WPA still works.


controllers/test/suite_e2e_test.go
```diff
@@ -15,7 +16,7 @@ import (
 func initTestConfig() *testConfigOptions {
 	return &testConfigOptions{
 		useExistingCluster: true,
-		crdVersion:         "v1beta1",
+		crdVersion:         "v1",
```
clamoriniere, line 19: does it means we are loosing `v1beta1` support?

clamoriniere (approved):

controllers/test/suite_e2e_test.go
juliogreff: No, this only affects tests. The testing implementation we use no longer supports v1beta1 (it's been deprecated for a bit), but since we support both, we don't have a problem.

PR #160, opened 2022-11-25
Add `Spec.ConvergeTowardsHighWatermark`

iksaif:
LW/HW are supposed to reduce oscillations, but when you have 1000 replicas with 20% between LW and HW you often end up with more than a hundred un-needed replicas.

Setting `Spec.ConvergeTowardsHighWatermark` will slowly downscale by `Spec.ReplicaScalingAbsoluteModulo` at a time, and only if the forecasted usage stays bellow the highwatermark.

The existing scale down cooldown mechanisms will be respected.

### DIFF
--- a/controllers/replica_calculator.go
+++ b/controllers/replica_calculator.go
@@ -286,7 +286,7 @@ func tryToConvergeToHw(logger logr.Logger, currentReplicas, currentReadyReplicas
 	setCondition(wpa, v1alpha1.WatermarkPodAutoscalerStatusConvergeToHighWatermark, corev1.ConditionTrue, convergeToHighWatermarkAllowedMessage, convergeToHighWatermarkAllowedReason)
 
 	// This should stay under HW
-	logger.Info("Trying to scale down to convert to HW", "replicaCount", replicaCount, "currentReadyReplicas", currentReadyReplicas, "currentReplicasAfterDownscale", currentReplicasAfterDownscale, "currentReadyReplicasAfterDownscale", currentReadyReplicasAfterDownscale, "adjustedUsageAfterDownscale", adjustedUsageAfterDownscale, "highMark", highMark.MilliValue())
+	logger.Info("Trying to scale down to converge to HW", "replicaCount", replicaCount, "currentReadyReplicas", currentReadyReplicas, "currentReplicasAfterDownscale", currentReplicasAfterDownscale, "currentReadyReplicasAfterDownscale", currentReadyReplicasAfterDownscale, "adjustedUsageAfterDownscale", adjustedUsageAfterDownscale, "highMark", highMark.MilliValue())
 	return currentReplicasAfterDownscale
 }
 
--- a/controllers/replica_calculator.go
+++ b/controllers/replica_calculator.go
@@ -26,10 +26,12 @@ import (
 )
 
 const (
-	aboveHighWatermarkAllowedMessage = "Allow upscaling if the value stays over the Watermark"
-	belowLowWatermarkAllowedMessage  = "Allow downscaling if the value stays under the Watermark"
-	aboveHighWatermarkReason         = "Value above High Watermark"
-	belowLowWatermarkReason          = "Value above Low Watermark"
+	aboveHighWatermarkAllowedMessage      = "Allow upscaling if the value stays over the Watermark"
+	belowLowWatermarkAllowedMessage       = "Allow downscaling if the value stays under the Watermark"
+	convergeToHighWatermarkAllowedMessage = "Allow downscaling to convert to the High Watermark"
+	aboveHighWatermarkReason              = "Value above High Watermark"
+	belowLowWatermarkReason               = "Value above Low Watermark"
+	convergeToHighWatermarkAllowedReason  = "Value between Watermarks"
 )
 
 // ReplicaCalculation is used to compute the scaling recommendation.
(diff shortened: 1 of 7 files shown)


controllers/replica_calculator.go
```diff
@@ -212,6 +214,89 @@ func (c *ReplicaCalculator) GetResourceReplicas(logger logr.Logger, target *auto
 	return ReplicaCalculation{replicaCount, utilizationQuantity, timestamp, int32(readyPodCount)}, nil
 }
 
+func Max(x, y int32) int32 {
```
clamoriniere, line 217: this function can be private and be rename to `maxInt32`

```suggestion
func maxInt32(x, y int32) int32 {
```

clamoriniere: I see lot of: 
`int32(math.Ceil(float64(..),float64(..))`, `int32(math.Mod(float64(..),float64(..))` or `int32(math.Floor(float64(..),float64(..))` in the following code change.

It would be nice to create a package `math32` to extract these functions and make the code more pleasant to read.

controllers/replica_calculator.go
```diff
@@ -212,6 +214,89 @@ func (c *ReplicaCalculator) GetResourceReplicas(logger logr.Logger, target *auto
...
+	setCondition(wpa, v1alpha1.WatermarkPodAutoscalerStatusBelowLowWatermark, corev1.ConditionFalse, belowLowWatermarkReason, belowLowWatermarkAllowedMessage)
+	setCondition(wpa, v1alpha1.WatermarkPodAutoscalerStatusConvergeToHighWatermark, corev1.ConditionFalse, convergeToHighWatermarkAllowedMessage, convergeToHighWatermarkAllowedReason)
+
+	// returning the currentReplicas instead of the count of healthy ones to be consistent with the upstream behavior.
+	return currentReplicas
+}
+
+// tryToConvergeToHw will try to make the replicaCount slowly converge to high watermark but make sure that it does
+//
+//	not trigger a downscale that would put usage above HW
```
clamoriniere, line 276:
```suggestion
// tryToConvergeToHw will try to make the replicaCount slowly converge to high watermark but make sure that it does
// not trigger a downscale that would put usage above HW
```

clamoriniere (commented):
few nits to make the code easier to read.

controllers/replica_calculator.go
iksaif: done

controllers/replica_calculator.go
```diff
@@ -212,62 +215,117 @@ func (c *ReplicaCalculator) GetResourceReplicas(logger logr.Logger, target *auto
...
+// not trigger a downscale that would put usage above HW
+func tryToConvergeToHw(logger logr.Logger, currentReplicas, currentReadyReplicas int32, wpa *v1alpha1.WatermarkPodAutoscaler, adjustedUsage float64, lowMark, highMark *resource.Quantity) (replicaCount int32) {
+
+	downScaleBy := *wpa.Spec.ReplicaScalingAbsoluteModulo
+	currentReplicasAfterDownscale := math32.Max(currentReplicas-downScaleBy, 1)
+	currentReadyReplicasAfterDownscale := math32.Max(currentReadyReplicas-downScaleBy, 0)
 
-		// returning the currentReplicas instead of the count of healthy ones to be consistent with the upstream behavior.
-		return currentReplicas, utilizationQuantity.MilliValue()
+	adjustedUsageAfterDownscale := math.Ceil(adjustedUsage * float64(currentReadyReplicas) / float64(currentReadyReplicasAfterDownscale))
```
CharlyF, line 276: if downscale by is > currentReadyReplicas, currentReadyReplicasAfterDownscale would be 0 or am I missing something?

controllers/replica_calculator.go
```diff
@@ -212,62 +215,117 @@ func (c *ReplicaCalculator) GetResourceReplicas(logger logr.Logger, target *auto
...
+		return getReplicaCountWithinBounds(currentReplicas, wpa)
 	}
 
-	restrictedScaling.With(labelsWithReason).Set(0)
+	setCondition(wpa, v1alpha1.WatermarkPodAutoscalerStatusAboveHighWatermark, corev1.ConditionFalse, aboveHighWatermarkReason, aboveHighWatermarkAllowedMessage)
+	setCondition(wpa, v1alpha1.WatermarkPodAutoscalerStatusBelowLowWatermark, corev1.ConditionFalse, belowLowWatermarkReason, belowLowWatermarkAllowedMessage)
+	setCondition(wpa, v1alpha1.WatermarkPodAutoscalerStatusConvergeToHighWatermark, corev1.ConditionTrue, convergeToHighWatermarkAllowedMessage, convergeToHighWatermarkAllowedReason)
+
+	// This should stay under HW
+	logger.Info("Trying to scale down to convert to HW", "replicaCount", replicaCount, "currentReadyReplicas", currentReadyReplicas, "currentReplicasAfterDownscale", currentReplicasAfterDownscale, "currentReadyReplicasAfterDownscale", currentReadyReplicasAfterDownscale, "adjustedUsageAfterDownscale", adjustedUsageAfterDownscale, "highMark", highMark.MilliValue())
```
CharlyF, line 289:
```suggestion
	logger.Info("Trying to scale down to converge to HW", "replicaCount", replicaCount, "currentReadyReplicas", currentReadyReplicas, "currentReplicasAfterDownscale", currentReplicasAfterDownscale, "currentReadyReplicasAfterDownscale", currentReadyReplicasAfterDownscale, "adjustedUsageAfterDownscale", adjustedUsageAfterDownscale, "highMark", highMark.MilliValue())
```

controllers/replica_calculator.go
```diff
@@ -212,62 +215,117 @@ func (c *ReplicaCalculator) GetResourceReplicas(logger logr.Logger, target *auto
...
+	if int64(adjustedUsageAfterDownscale) > highMark.MilliValue() {
+		// This would result in a downscale, give up
+		logger.Info("Scaling down would likely make usage go above HW", "replicaCount", replicaCount, "currentReadyReplicas", currentReadyReplicas, "currentReadyReplicasAfterDownscale", currentReadyReplicasAfterDownscale, "adjustedUsageAfterDownscale", adjustedUsageAfterDownscale, "highMark", highMark.MilliValue())
+		return getReplicaCountWithinBounds(currentReplicas, wpa)
 	}
 
-	restrictedScaling.With(labelsWithReason).Set(0)
+	setCondition(wpa, v1alpha1.WatermarkPodAutoscalerStatusAboveHighWatermark, corev1.ConditionFalse, aboveHighWatermarkReason, aboveHighWatermarkAllowedMessage)
+	setCondition(wpa, v1alpha1.WatermarkPodAutoscalerStatusBelowLowWatermark, corev1.ConditionFalse, belowLowWatermarkReason, belowLowWatermarkAllowedMessage)
+	setCondition(wpa, v1alpha1.WatermarkPodAutoscalerStatusConvergeToHighWatermark, corev1.ConditionTrue, convergeToHighWatermarkAllowedMessage, convergeToHighWatermarkAllowedReason)
```
CharlyF, line 286: I think if we are in stable regime (i.e. between bounds) and we optimize to converge towards a watermark we should not affect the scaling windows, because we are not scaling as a result of the external signal but from our internal logic. 

I think we should improve the logic in the [`shouldScale`](https://github.com/DataDog/watermarkpodautoscaler/blob/main/controllers/watermarkpodautoscaler_controller.go#L356-L414) method to take this into account both for:
- Stable regime: Circumvent (?) `shouldScale` to have the adjustedUsage converge towards the HW with no restrictions.
- Unstable regime: A previous optimization should not prevent us from scaling how we would have if we had not used this feature. Thus, we do not affect the specific parts of the status used for the window calculation.

CharlyF (changes requested):
I am a bit hesitant to introduce logic that allows scaling events when we are in stable regime because I believe it can be used as a source of truth for users.
However, I see the value of this feature, and I do not have a better way to handle your use case at the moment. In the long term, I was thinking about how having dynamic watermarks could help?

I left some suggestions, especially pertaining to the way the optimization could affect the regular lifecycle (i.e. blocking a legitimate scaling event because we just optimized).
Further, I can imagine a use case for the opposite version of this feature (upscaling for the adjustedUsage to converge towards the LW as a way to pre-warm caches maybe). If you think this is legitimate, I would suggest changing `ConvergeTowardsHighWatermark` into `ConvergeTowardsWatermark` as to take a string with `low`, `high` or `disabled`?

iksaif:
> I am a bit hesitant to introduce logic that allows scaling events when we are in stable regime because I believe it can be used as a source of truth for users.

We do have scaling events in stable regime all the time, just because we often have long downscale cooldowns for example

> However, I see the value of this feature, and I do not have a better way to handle your use case at the moment. In the long term, I was thinking about how having dynamic watermarks could help?

I though about dynamic watermarks, but I could not come up with a good user interface for them, maybe somebody else will be able to ?
 
> I left some suggestions, especially pertaining to the way the optimization could affect the regular lifecycle (i.e. blocking a legitimate scaling event because we just optimized).

Blocking a downscale because we did an optimisation downscale is probably not a very big issue no ?

> Further, I can imagine a use case for the opposite version of this feature (upscaling for the adjustedUsage to converge towards the LW as a way to pre-warm caches maybe). If you think this is legitimate, I would suggest changing `ConvergeTowardsHighWatermark` into `ConvergeTowardsWatermark` as to take a string with `low`, `high` or `disabled`?

Maybe we should ask some people if they would use such feature ?

controllers/replica_calculator.go
iksaif: That's totally correct, we never want `currentReadyReplicasAfterDownscale` to be 0 anyway, the Max should be `1` too

iksaif:
👋 @CharlyF - what do you think about `ConvergeTowardsWatermark` vs `ConvergeTowardsHighWatermark` ?

iksaif:
@CharlyF is taking over

PR #161, opened 2022-11-28
Add ARM support

clamoriniere:
### What does this PR do?

* migrate to go-licenses to generate 3rd party licenses csv
* update release CI pipeline to also build ARM image
* Update wpa kubectl plugin release process to also build ARM bin
* Add kubebuilder-tools tools for e2e tests.
* Update golangci-lint
* Update CODEOWNERS file to add #container-ecosystems

### Motivation

be able to use `kubectl wpa` on an ARM laptop. be able to test locally on an ARM laptop

### Additional Notes

### Describe your test plan


.github/workflows/release.yaml
```diff
@@ -1,51 +1,164 @@
...
-      - 
-        name: Get Latest Release
+          go-version: 1.18
+      - name: Get Latest Release
         id: latest_version
         uses: pozetroninc/github-action-get-latest-release@master
         with:
             owner: ${{ github.repository_owner }}
-            repo: watermarkpodautoscaler
+            repo: extendeddaemonset
```
CharlyF, line 34: s/extendeddaemonset/watermarkpodautoscaler/

LICENSE-3rdparty.csv
```diff
@@ -1,888 +1,83 @@
 Component,Origin,License
-core,"github.com/frapposelli/wwhrd",MIT
```
CharlyF, line 2: There is a significant diff in the licenses, are we not losing coverage?

CharlyF (changes requested):
can you add the return line at the end of most files?

.github/workflows/release.yaml
clamoriniere: nice catch

LICENSE-3rdparty.csv
clamoriniere: no, I forgot to add a comment. 

it is only because the `go-licenses` program only the package really used by the agent, not was is vendor

(btw: I'm planning another PR to remove the `vendoring` logic in the makefile)

CharlyF (approved):
thank you!!

PR #171, opened 2023-02-27
fix(readiness_delay_s): change condition of ignore pod

nicolasgodefroy:
### What does this PR do?

Definition : 
    
* readinessDelaySeconds: Specifies how much time replicas need to be running for, prior to be taken into account in the scaling decisions.

Change the condition if the pod is ready and the readinessDelaySeconds is after the last transition time of the pod.

### Motivation

Fix the readiness_delay_sec condition.

### Additional Notes

Fix init with make intall-tools

### Describe your test plan

* Create a pod with WPA and a readiness_delay_sec 120s
* Launch a load test.
* Check that the wpa service wait the readiness_delay_sec before scaling up the pod.
* Validate the test plan.


CharlyF (approved):
Great find, thank you very much for opening this up.

PR #172, opened 2023-02-27
Adding support for delayScale with multi metrics

CharlyF:
### What does this PR do?

The current feature to delay scaling decision only supports one metric.
This introduces the logic necessary to support multiple metrics.
Note that for now the logic will use `OR`, so not all the metrics need to be beyond their watermarks for the duration of the delay. Further, if the union of the metrics's time beyond their respective watermarks goes longer than the configured delay, this will also allow for a scaling event to occur.

We would consider using `AND` in order to only scale if *all* metrics are beyond their respective watermarks, but it'd require a larger change with refactoring of the Status to report all the metrics's positions.

### Motivation

Feature Request

### Additional Notes

The best way to monitor this is by using the log: `Will not scale: value has not been out of bounds for long enough` that has the `time_left` as an argument. 
Alternatively, you can check the conditions of your WPA for `AbleToScale` with status `ReadyForScale` set to `True` (unless in Dry-Run).

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ In the following example, we can see that the recommended number of replicas is 
 * **Scaling Delay**
 <a name="delay"></a>
 
-In order to avoid scaling from bursts you can use the following features: `downscaleDelayBelowWatermarkSeconds` and/or `upscaleDelayAboveWatermarkSeconds`. These options are specified as integers. The metric has to remain above or under its respective watermark for the configured duration.
+In order to avoid scaling from bursts you can use the following features: `downscaleDelayBelowWatermarkSeconds` and/or `upscaleDelayAboveWatermarkSeconds`. These options are specified as integers. The metric(s) have to remain above or under its/their respective watermark for the configured duration.
 You can keep track of how much time is left in the status of the WPA:
 
 ```
@@ -188,7 +188,9 @@ Or in the logs of the controller:
 {"level":"info","ts":1668481092517.446,"logger":"controllers.WatermarkPodAutoscaler","msg":"Will not scale: value has not been out of bounds for long enough","watermarkpodautoscaler":"datadog/example-watermarkpodautoscaler","wpa_name":"example-watermarkpodautoscaler","wpa_namespace":"datadog","time_left":3209}
 ```
 
-**Note:** This feauture does not support using multiple metrics. 
+**Note:** If you are using multiple metrics with this feature, the above/below condition is considered using the `OR` of the metrics.
+
+e.g. if you have a 60s upscaleDelay with 2 metrics (M1 and M2), M1 stays above it's high watermark for 40s `[t0; t40]` and the M2 one goes above it's high watermark for 30s while overlapping with M1 during it's last 10s: `[t30; t60]`, this will validate the upscaleDelay condition and allow for an upscaling event.
 
 * **Precedence**
 <a name="precedence"></a>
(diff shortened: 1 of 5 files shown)


controllers/replica_calculator.go
```diff
@@ -38,6 +38,12 @@ type ReplicaCalculation struct {
 	utilization   int64
 	timestamp     time.Time
 	readyReplicas int32
+	pos           MetricPosition
+}
+
+type MetricPosition struct {
```
clamoriniere, line 44: I think this struct can be private

```suggestion
type metricPosition struct {
```

README.md
```diff
@@ -171,7 +171,7 @@ In the following example, we can see that the recommended number of replicas is
 * **Scaling Delay**
 <a name="delay"></a>
 
-In order to avoid scaling from bursts you can use the following features: `downscaleDelayBelowWatermarkSeconds` and/or `upscaleDelayAboveWatermarkSeconds`. These options are specified as integers. The metric has to remain above or under its respective watermark for the configured duration.
+In order to avoid scaling from bursts you can use the following features: `downscaleDelayBelowWatermarkSeconds` and/or `upscaleDelayAboveWatermarkSeconds`. These options are specified as integers. The metric(s) have to remain above or under its/their respective watermark for the configured duration.
```
kayayarai, line 174: This was better phrased before, in terms of translatability :)
```suggestion
In order to avoid scaling from bursts you can use the following features: `downscaleDelayBelowWatermarkSeconds` and/or `upscaleDelayAboveWatermarkSeconds`. These options are specified as integers. Each metric has to remain above or under its respective watermark for the configured duration.
```

README.md
```diff
@@ -188,7 +188,9 @@ Or in the logs of the controller:
 {"level":"info","ts":1668481092517.446,"logger":"controllers.WatermarkPodAutoscaler","msg":"Will not scale: value has not been out of bounds for long enough","watermarkpodautoscaler":"datadog/example-watermarkpodautoscaler","wpa_name":"example-watermarkpodautoscaler","wpa_namespace":"datadog","time_left":3209}
 ```
 
-**Note:** This feauture does not support using multiple metrics. 
+**Note:** If you are using multiple metrics with this feature, the above/below condition is considered using the `OR` of the metrics.
+
+e.g. if you have a 60s upscaleDelay with 2 metrics (M1 and M2), M1 stays above it's high watermark for 40s `[t0; t40]` and the M2 one goes above it's high watermark for 30s while overlapping with M1 during it's last 10s: `[t30; t60]`, this will validate the upscaleDelay condition and allow for an upscaling event.
```
kayayarai, line 193:
```suggestion
For example, suppose you have a 60 second `upscaleDelay` with two metrics, M1 and M2. If M1 stays above its high watermark for 40 seconds `[t0; t40]`, and the M2 one goes above its high watermark for 30 seconds, overlapping with M1 during its last 10 seconds, `[t30; t60]`, this validates the `upscaleDelay` condition and allows for an upscaling event.
```

README.md
```diff
@@ -216,7 +218,7 @@ If all the conditions are met, the controller will scale the targeted object in
 
 ## Limitations
 
-- Only officially supports one metric per WPA.
+- Only officially supports one metric per WPA: While the logic supports multiple metrics and will apply the greatest recommendation of all metrics, the status needs some refactoring to reflect this insight.
```
kayayarai, line 221:
```suggestion
- Only officially supports one metric per WPA. While the logic supports multiple metrics and applies the greatest recommendation of all metrics, the status needs some refactoring to reflect this insight.
```

kayayarai (approved):
Suggested docs spelling and style improvements

README.md
CharlyF: With `Each` my understanding is that we use `AND` and need all the metrics to stay above/below for the duration whereas we only need the union of the metrics to be above/below.

clamoriniere (approved):
🚀

PR #179, opened 2023-04-10
Fix arm release & Update tooling

CharlyF:
### What does this PR do?

- Update the Operator-sdk
- Update go
- Update tooling (yq, controller-gen)
- Generate manifests
- Update logs for the convergence logic

### Motivation

Preparing the release of 0.6.0

### Additional Notes

The update of `hack/patch-crds.sh` drops the toleration sed - we need to check that this does not break the upgrade path from the older CRD where the toleration was a different type.

### Describe your test plan

Cut 0.6.0-rc.1, test the deploy and the upgrade of the CRD.


config/crd/kustomization.yaml
```diff
@@ -2,7 +2,7 @@
 # since it depends on service name and namespace that are out of this kustomize package.
 # It should be run by config/default
 resources:
-- bases/v1beta1/datadoghq.com_watermarkpodautoscalers.yaml
+- bases/v1/datadoghq.com_watermarkpodautoscalers.yaml
```
clamoriniere, line 5: is it expected?

clamoriniere, line 5: this might require some doc in the release to say that K8s >= 1.16 on is supported

config/manifests/bases/watermarkpodautoscaler.clusterserviceversion.yaml
```diff
@@ -39,128 +39,21 @@ metadata:
...
 spec:
   apiservicedefinitions: {}
-  customresourcedefinitions:
-    owned:
-    - description: WatermarkPodAutoscaler is the Schema for the watermarkpodautoscalers API
-      displayName: Watermark Pod Autoscaler
-      kind: WatermarkPodAutoscaler
-      name: watermarkpodautoscalers.datadoghq.com
-      version: v1alpha1
+  customresourcedefinitions: {}
```
clamoriniere, line 48: it is weird that this file is now almost empty

clamoriniere (commented):
the golang version can be also updated in 
* [go.mod](https://github.com/DataDog/watermarkpodautoscaler/blob/main/go.mod#L3)
* and in the different github workflow like https://github.com/DataDog/watermarkpodautoscaler/blob/main/.github/workflows/build.yaml#L22

config/manifests/bases/watermarkpodautoscaler.clusterserviceversion.yaml
CharlyF: agreed, I can look into this when I cut 0.6.0-rc.X, this is just so I can cut it

clamoriniere (approved):

PR #188, opened 2023-07-10
Add scaling events count to kubectl output

khewonc:
### What does this PR do?

Adds a total count of scaling events to `kubectl describe` and `kubectl get -o wide` commands:

get output:
```shell   
$ kubectl get wpa -o wide
NAME   SCALING ACTIVE   CONDITION     CONDITION STATE   VALUE   HIGH WATERMARK   LOW WATERMARK   AGE   MIN REPLICAS   MAX REPLICAS   DRY-RUN   LAST SCALE   SCALE COUNT
wpa    True             AbleToScale   False             10      15               1               10s   1              2              False     6s           1                
```

describe output:
```
$ kubectl describe wpa wpa
Name:         wpa
Namespace:    default
[...]
  Last Scale Time:            2023-07-10T21:50:28Z
  Scaling Events Count:       2
```

### Motivation

If it's not possible to look at metrics or parse through logs to confirm scaling events are still occurring during an outage, an increasing number of `Scaling Events Count` would indicate that the WPA was still able to scale up/down

### Additional Notes

The output of `kubectl get` already includes a lot of information and since the purpose of this counter is for use in the (hopefully) unlikely case of unavailable metrics, the `SCALE COUNT` is only available in the `kubectl get -o wide` view. It is always available in a `kubectl describe` output

### Describe your test plan

Scale several WPAs multiple times and ensure the scale count total is correct for each WPA

### DIFF
--- a/controllers/watermarkpodautoscaler_controller.go
+++ b/controllers/watermarkpodautoscaler_controller.go
@@ -460,6 +460,7 @@ func setStatus(wpa *datadoghqv1alpha1.WatermarkPodAutoscaler, currentReplicas, d
 	if rescale {
 		now := metav1.NewTime(time.Now())
 		wpa.Status.LastScaleTime = &now
+		wpa.Status.ScalingEventsCount += 1
 	}
 }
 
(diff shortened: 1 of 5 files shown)


controllers/watermarkpodautoscaler_controller.go
```diff
@@ -460,6 +460,7 @@ func setStatus(wpa *datadoghqv1alpha1.WatermarkPodAutoscaler, currentReplicas, d
 	if rescale {
 		now := metav1.NewTime(time.Now())
 		wpa.Status.LastScaleTime = &now
+		wpa.Status.ScalingEventsCount += 1
```
levan-m, line 463: Do we ever reset this to 0 or does it keep increasing? Should it be reset when WPA is modified?

CharlyF: I agree that we should revisit this. Though it should be applied to the entire status. I will add a card to review whether we should improve on it.

PR #197, opened 2023-10-10
Introducing option to avoid scaling if value is 0 by default

CharlyF:
### What does this PR do?

The Metrics Providers can return `0` with a 200 in certain cases, which can lead to really bad outcomes (maximum downscaling effect), especially if the filters (downscaleForbiddenWindow, scaleDownLimit...) are not used properly.
For most metrics used, `0` is not an acceptable value (memory, cpu). However, there are cases where it does make sense (requests, workqueue).
Because of the impact it can have, this will prevent 0 from being tolerated by default with a log in case users are not aware of the option, they will just see their workloads not downscale as expected with the controller logging why.

### Motivation

Better resilience.

### Describe your test plan

Pick a metric that is worth `0`, verify that the WPA does nothing by default and set the option to see it downscale the workload.


clamoriniere (commented):
the logic is good. 
But I think we are implemented a workaround, but we are not fixing the root cause issue.

controllers/replica_calculator.go
```diff
@@ -138,12 +138,17 @@ func (c *ReplicaCalculator) GetExternalMetricReplicas(logger logr.Logger, target
...
 	}
-	logger.Info("Metrics from the External Metrics Provider", "metrics", metrics)
+	logger.V(2).Info("Metrics from the External Metrics Provider", "metrics", metrics)
 
 	var sum int64
 	for _, val := range metrics {
 		sum += val
 	}
+	if sum == 0 && !wpa.Spec.TolerateZero {
+		logger.V(1).Info("Retrieved 0 from the External Metrics Provider which is not tolerated by default, use Spec.TolerateZero to enable")
```
clamoriniere, line 148: IMO it can be a `Warning`

PR #204, opened 2024-03-08
Make generating NotScale events configurable; use logs when disabled

levipe01:
### What does this PR do?

Add skipNotScalingEvents config option in WPA to disable NotScaling Kubernetes events. When disabled this will generate a log instead and update the `Status` of the WPA. 

### Motivation

Eliminate WPA error logs that are the result of the Kubernetes API server rate limiting WPA event writes

### Additional Notes

- When updating the `Status` we use the `ScalingActive` Condition Type. This is initially set [here](https://github.com/DataDog/watermarkpodautoscaler/blob/levipe01/notscale-events-config-option/controllers/watermarkpodautoscaler_controller.go#L687) and we override the condition & message to show that the replicas are correctly calculated, however we are choosing not to scale.
- Leaving the default behavior to continue to write Kubernetes Events in the event that 3rd party users are currently monitoring on these.

### Describe your test plan

- Enabling this feature should result in the Kubernetes events with a message of `Decided not to scale` to appear as logs instead.
- When describing the WPA object you should see the `Status` update with a message like the below when not scaling.
```
Status:
  Conditions:
    Last Transition Time:  2024-03-21T21:46:18Z
    Message:               the WPA was able to successfully calculate a replica count and decided not to scale Deployment/datadog/php-apache to 9 (last scale time was 2024-03-21 21:27:48 +0000 UTC )
    Reason:                NotScaling
    Status:                True
    Type:                  ScalingActive
```

### DIFF
--- a/controllers/watermarkpodautoscaler_controller.go
+++ b/controllers/watermarkpodautoscaler_controller.go
@@ -70,6 +70,10 @@ const (
 	desiredCountAcceptable = "the desired count is within the acceptable range"
 )
 
+type Options struct {
+	SkipNotScalingEvents bool
+}
+
 // WatermarkPodAutoscalerReconciler reconciles a WatermarkPodAutoscaler object
 type WatermarkPodAutoscalerReconciler struct {
 	// This client, initialized using mgr.Client() above, is a split client
@@ -82,6 +86,7 @@ type WatermarkPodAutoscalerReconciler struct {
 	syncPeriod    time.Duration
 	eventRecorder record.EventRecorder
 	replicaCalc   ReplicaCalculatorItf
+	Options       Options
 }
 
 // +kubebuilder:rbac:groups=apps;extensions,resources=deployments/finalizers,resourceNames=watermarkpodautoscalers,verbs=update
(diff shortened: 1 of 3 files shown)


levan-m (dismissed):

vboulineau:
The code LGTM but I don't think we should emit an event ever. Either we emit a debug log or we just update the `Status` with last eval timestamp or we just produce a metric with `1` value every time we eval.

clamoriniere (approved):

controllers/watermarkpodautoscaler_controller.go
```diff
@@ -373,7 +378,13 @@ func (r *WatermarkPodAutoscalerReconciler) reconcileWPA(ctx context.Context, log
 
 		logger.Info("Successful rescale", "currentReplicas", currentReplicas, "desiredReplicas", desiredReplicas, "rescaleReason", rescaleReason)
 	} else {
-		r.eventRecorder.Eventf(wpa, corev1.EventTypeNormal, datadoghqv1alpha1.ReasonNotScaling, fmt.Sprintf("Decided not to scale %s to %d (last scale time was %v )", reference, desiredReplicas, wpa.Status.LastScaleTime))
+		if r.Options.SkipNotScalingEvents {
+			setCondition(wpa, autoscalingv2.ScalingActive, corev1.ConditionTrue, datadoghqv1alpha1.ConditionReasonNotScaling, "the WPA was able to successfully calculate a replica count and decided not to scale %s to %d (last scale time was %v )", reference, desiredReplicas, wpa.Status.LastScaleTime)
+			logger.Info(fmt.Sprintf("Decided not to scale %s to %d (last scale time was %v )", reference, desiredReplicas, wpa.Status.LastScaleTime))
```
vboulineau, line 383: I think having the info once in the status is enough. The WPA is already a big logger, could be a `debug` log if you want to keep it.

vboulineau (approved):

PR #219, opened 2024-07-10
tool: add a script to centralize golang version update

clamoriniere:
### What does this PR do?

the New script is taking the version from the `go.mod` file thank to the `go mod edit --json` command ouput.
The idea is to first update the golang version in the `go.mod` file, then run the command `make update-golang` that will patch all the files that contains the golang version.

commits:
- **apply linter on .github/workflow files**
- **Add script to centralise golang version update**

### Motivation

To not forget to update the golang version in one of the many file that have the golang version in parameter.
we needed to have a solution to centralise the update of the golang version, but also to verify that all file are in sync

### Additional Notes

### Describe your test plan


hack/install-jq.sh
```diff
@@ -0,0 +1,38 @@
...
+set -o pipefail
+
+SCRIPTS_DIR="$(dirname "$0")"
+# Provides $OS,$ARCH,$PLATFORM,$ROOT variables
+source "$SCRIPTS_DIR/install-common.sh"
+
+cleanup() {
+  rm -rf "$WORK_DIR"
+}
+trap "cleanup" EXIT SIGINT
```
L3n41c, line 14: Is this still needed as it is already done inside `install-common.sh` ?
https://github.com/DataDog/watermarkpodautoscaler/blob/02ebc32ba5e2eaa8600c77f154c915efd0652176/hack/install-common.sh#L46-L52

L3n41c (approved):

hack/install-jq.sh
clamoriniere: no,
I copy pasted the install-yq.sh script

I will do the change in both

clamoriniere:
@L3n41c PR updated and ready for a second path.

PR #221, opened 2024-10-03
[CONTINT-4412] Upgrade `k8s` dependencies

L3n41c:
### What does this PR do?

Update kubernetes and related dependencies.

### Motivation

We need to update `watermarkpodautoscaler` to be able to [update in `datadog-agent`](https://github.com/DataDog/datadog-agent/blob/6d5f968e0dd5ee4a7504d780e398f876d5c8de27/go.mod#L166).

### Additional Notes

### Describe your test plan


go.mod
```diff
@@ -1,159 +1,116 @@
...
 
-go 1.21
+go 1.22.0
+
+toolchain go1.22.7
 
 require (
-	github.com/DataDog/datadog-operator v1.3.0
-	github.com/go-logr/logr v1.3.0
+	github.com/DataDog/datadog-operator v0.7.1-0.20241003091406-6e9e14d97438
```
clamoriniere, line 8: does this tag correspond to main?
can we use `v1.8.0` ?

main.go
```diff
@@ -106,22 +107,23 @@ func main() {
...
+		Scheme: scheme,
+		Metrics: metricsserver.Options{
+			BindAddress: fmt.Sprintf("%s:%d", host, metricsPort),
+		},
 		LeaderElection:             enableLeaderElection,
 		LeaderElectionID:           "watermarkpodautoscaler-lock",
 		LeaderElectionResourceLock: leaderElectionResourceLock,
 		HealthProbeBindAddress:     fmt.Sprintf("%s:%d", host, healthPort),
-		SyncPeriod:                 &syncDuration,
+		// SyncPeriod:                 &syncDuration,
```
clamoriniere, line 126: does it means that this option was removed?

clamoriniere (commented):
few comments

go.mod
L3n41c: This tag corresponds to the opened PR DataDog/datadog-operator#1445 because I preferred to have all the dependencies aligned and avoid having a bunch of different versions for a given dependency (like `k8s.io` modules) depending on the path that pulls it.

So, this will need to be updated to a `main` commit as soon as DataDog/datadog-operator#1445 will be merged.

main.go
```diff
@@ -106,22 +107,23 @@ func main() {
 	}
 	version.PrintVersionLogs(setupLog)
 
-	syncDuration := time.Duration(syncPeriodSeconds) * time.Second
+	// syncDuration := time.Duration(syncPeriodSeconds) * time.Second
```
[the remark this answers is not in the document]
L3n41c: I put this option back.

main.go
L3n41c: This option was moved. I’ve updated this piece of code with the new option location.

iksaif (approved):

clamoriniere (approved):

PR #230, opened 2024-11-07
recommender: fix instrumentation and add tests

iksaif:
Fix recommender metrics ("recommender" != "client")

### DIFF
--- a/controllers/datadoghq/metrics.go
+++ b/controllers/datadoghq/metrics.go
@@ -29,7 +29,10 @@ const (
 	transitionPromLabel        = "transition"
 	lifecycleStatus            = "lifecycle_status"
 	monitorName                = "monitor_name"
-	monitorNamespace           = "monotor_namespace"
+	monitorNamespace           = "monitor_namespace"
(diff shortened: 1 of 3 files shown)


controllers/datadoghq/metrics.go
```diff
@@ -29,7 +29,10 @@ const (
 	transitionPromLabel        = "transition"
 	lifecycleStatus            = "lifecycle_status"
 	monitorName                = "monitor_name"
-	monitorNamespace           = "monotor_namespace"
+	monitorNamespace           = "monitor_namespace"
+	clientPromLabel            = "client"
```
vboulineau, line 33: The `recommender` name was perhaps more meaningful. As it's a not generic HTTP Client but only dedicated to call recommenders.

iksaif: sure, but there metrics could be used for any http call (now that they are in "metrics") - and client contains the url of the recommender when it's a recommender

vboulineau (approved):

PR #240, opened 2024-12-03
Add option to change QPS limit of Kubernetes client

vboulineau:
### What does this PR do?

Add option to change QPS limit of Kubernetes client

### Motivation

Allow WPA to scale to multiple thousands of WPA

### Describe your test plan

Visible in the number of external queries not being bounded to the default 20 / second.

### DIFF
--- a/main.go
+++ b/main.go
@@ -59,6 +59,7 @@ func main() {
 	var logTimestampFormat string
 	var syncPeriodSeconds int
 	var clientTimeoutDuration time.Duration
+	var clientQPSLimit float64
 	var ddProfilingEnabled bool
 	var workers int
 	var skipNotScalingEvents bool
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -71,6 +72,7 @@ func main() {
 	flag.StringVar(&logTimestampFormat, "log-timestamp-format", "millis", "log timestamp format ('millis', 'nanos', 'epoch', 'rfc3339' or 'rfc3339nano')")
 	flag.IntVar(&syncPeriodSeconds, "syncPeriodSeconds", 60*60, "The informers resync period in seconds")                                            // default 1 hour
 	flag.DurationVar(&clientTimeoutDuration, "client-timeout", 0, "The maximum length of time to wait before giving up on a kube-apiserver request") // is set to 0, keep default
+	flag.Float64Var(&clientQPSLimit, "client-qps", 0, "QPS Limit for the Kubernetes client (default 20 qps)")
```
clamoriniere, line 75: should we put 20 in the default value?
or maybe `-1` to clearly explain that `-1` means we use the default from the kubernetes client

vboulineau: `0` plays the role if `-1` as it defaults to default (`5` in `client-go` and `20` in `controller-runtime`)

clamoriniere: 0 is interpreted as unset in the logic that you have added or in

clamoriniere (approved):

PR #242, opened 2024-12-04
[controllers/wpa] Remove dependency from datadog-operator by using Unstructured

davidor:
### What does this PR do?

Removes the dependency on the datadog-operator project.

The WPA only depends on the datadog-operator to get the DatadogMonitor CRD for the "Lifecycle control" feature introduced in this PR: https://github.com/DataDog/watermarkpodautoscaler/pull/202 

However, the WPA only needs to access one specific field within the monitor. Therefore, using Kubernetes "unstructured" objects to interact with this CRD is a feasible solution.

We'd like to avoid the dependency from the datadog-operator because dependencies between projects that require kubernetes libraries can be problematic. Sometimes, in order to upgrade the kubernetes dependencies in one project, we need to upgrade the other project first, and we'd like to avoid this situation so that each project can be upgraded independently.

Until now the WPA depended on the datadog-operator, and also EDS indirectly. That's no longer the case.

### Describe your test plan

Test that the lifecycle control feature still works as described in the original PR https://github.com/DataDog/watermarkpodautoscaler/pull/202

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,6 @@ toolchain go1.22.7
 
 require (
 	github.com/DataDog/agent-payload/v5 v5.0.135-0.20241016074204-82c61e7189e0
-	github.com/DataDog/datadog-operator v0.7.1-0.20241104215218-3718019cd024
 	github.com/go-logr/logr v1.4.2
 	github.com/onsi/ginkgo v1.16.5
 	github.com/onsi/gomega v1.34.1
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -35,10 +34,10 @@ require (
 
 require (
 	github.com/Azure/go-ansiterm v0.0.0-20210617225240-d185dfc1b5a1 // indirect
-	github.com/DataDog/datadog-api-client-go/v2 v2.27.0 // indirect
+	github.com/DataDog/datadog-agent/pkg/obfuscate v0.59.0-rc.5 // indirect
+	github.com/DataDog/datadog-agent/pkg/remoteconfig/state v0.59.0-rc.5 // indirect
```
clamoriniere, line 38: do we need this datadog-agent package in WPA?

davidor: These 2 are unrelated to my changes. They were added when I ran `go mod tidy`, so it was either out of sync or there's an issue in my setup.

Both of them are dependencies of `gopkg.in/DataDog/dd-trace-go.v1` (already there before my PR).

```
go mod why -m github.com/DataDog/datadog-agent/pkg/obfuscate
# github.com/DataDog/datadog-agent/pkg/obfuscate
github.com/DataDog/watermarkpodautoscaler
gopkg.in/DataDog/dd-trace-go.v1/profiler
gopkg.in/DataDog/dd-trace-go.v1/profiler.test
gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer
github.com/DataDog/datadog-agent/pkg/obfuscate
```

```
go mod why -m github.com/DataDog/datadog-agent/pkg/remoteconfig/state
# github.com/DataDog/datadog-agent/pkg/remoteconfig/state
github.com/DataDog/watermarkpodautoscaler
gopkg.in/DataDog/dd-trace-go.v1/profiler
gopkg.in/DataDog/dd-trace-go.v1/profiler.test
gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer
github.com/DataDog/datadog-agent/pkg/remoteconfig/state
```

In any cases these are specific packages in the agent so it shouldn't cause any issues when upgrading k8s libs.

clamoriniere: ok thanks for looking at it. 
I was scared that we have yet another cross project dependency.

clamoriniere (approved):

L3n41c (approved):

PR #252, opened 2024-12-18
Add TLS configuration to Recommender http client

masterzen:
### What does this PR do?

This PR adds a configuration section for TLS in the `RecommenderSpec`:

```
// TLSConfig specifies the Recommender http client TLS configuration
// +k8s:openapi-gen=true
type TLSConfig struct {
	// CAFile is a path to a CA certificate
	// +optional
	CAFile string `json:"caFile,omitempty"`

	// CertFile is a path to a client Cert
	// +optional
	CertFile string `json:"certFile,omitempty"`

	// Keyfile is a path to the client certificate key (mandatory if CertFile is set)
	// +optional
	KeyFile string `json:"keyFile,omitempty"`

	// ServerName is a settings to activate TLS SNI

	ServerName string `json:"serverName,omitempty"`

	// InsecureSkipVerify when true disable verifying server certificate
	InsecureSkipVerify bool `json:"insecureSkipVerify,omitempty"`

	// MinVersion, minimum TLS version, defaults to Go default which is TLS1.2
	// +optional
	MinVersion string `json:"minVersion,omitempty"`

	// MinVersion, maximum TLS version, defaults to Go default which is TLS1.3
	// +optional
	MaxVersion string `json:"maxVersion,omitempty"`
}
```

And sets the HTTP client used internally when contacting the Recommender API according to those settings. For instance this allows to set client certificate or the server Root CA for the server certificate verification.

### Motivation

There was no way to setup proper mTLS with the current way and thus in an environment where mTLS is mandatory, the Recommender API was failing with TLS communication errors.

### Additional Notes

Note that the TLS settings are per RecommenderSpec to be able to adapt to various environment. To support environments where certificates have a short lifetime and are rotated frequently, the http client is set in the PR to reload the TLS client certificate from disk at every HTTP request. This is probably fine in this context as the recommender API is not used very frequently. If that's not something desirable we can implement a modification check to reload them from disk only when they have changed.

### Describe your test plan

This has been tested successfully on chillpenguin with emissary version 1.39.0-AD1 and the current admission hooks that injects this version and support file mode.
It was observed that the recommender HTTPS connection was using client certificate authentication and that it was successfully accepted by the server.

image


controllers/datadoghq/tls.go
```diff
@@ -0,0 +1,92 @@
...
+		if err == nil {
+			rootCA = x509.NewCertPool()
+			if !rootCA.AppendCertsFromPEM(caPEM) {
+				return nil, fmt.Errorf("failed to append root CA to pool")
+			}
+		}
+	}
+
+	// reload client certificate on each client connection
+	getClientCertificate := func(*tls.CertificateRequestInfo) (*tls.Certificate, error) {
```
iksaif, line 66: Maybe we should have a cache here, because the WPA is going to open many connections, and that's probably going to affect performances.

We could take the expiration time into account and use an expiring cache no ?

apis/datadoghq/v1alpha1/watermarkpodautoscaler_types.go
```diff
@@ -132,6 +132,37 @@ type WatermarkPodAutoscalerSpec struct {
...
+	// ServerName is a settings to activate TLS SNI
+
+	ServerName string `json:"serverName,omitempty"`
+
+	// InsecureSkipVerify when true disable verifying server certificate
+	InsecureSkipVerify bool `json:"insecureSkipVerify,omitempty"`
+
+	// MinVersion, minimum TLS version, defaults to Go default which is TLS1.2
+	// +optional
+	MinVersion string `json:"minVersion,omitempty"`
```
vboulineau, line 159: This is a new feature, we don't need to support anything else than 1.3. Similarly, if we dont have a use case for `ServerName` right now, I would drop it and potentially add later if needed.

controllers/datadoghq/tls.go
masterzen: There's only very few HTTP request per recommender (I think it's one request per minute). Since the certificate can be different per recommender URL, caching might not have much effect.
What I wanted to add was to check if the files changed before reloading, do you think that'd be enough or do you think I should a proper in-memory caching?

controllers/datadoghq/tls.go
```diff
@@ -0,0 +1,232 @@
...
+	return func(*tls.CertificateRequestInfo) (*tls.Certificate, error) {
+		now := time.Now()
+		c.mu.RLock()
+		entry, ok := c.cache[certFile]
+		c.mu.RUnlock()
+
+		// there's a small race condition here, but it does no harm except
+		// possibly reloading the certificate from disk more than once at a given time
+		if !ok || entry.isExpired(now) || entry.isCertificateExpired(now) {
+			certificate, err := retryLoadingX09Keypair(5, 50*time.Millisecond, certFile, keyFile)
```
iksaif, line 61: Could we kill the WPA here if we continusouly reload a bad certificate a lot ?

controllers/datadoghq/tls.go
```diff
@@ -0,0 +1,232 @@
...
+	certificateCacheTimeout    = 10 * time.Minute
+	certificateCacheLRUTimeout = 10 * time.Minute
+)
+
+type tlsTransport struct {
+	wrappedTransport *http.Transport
+}
+
+// tlsCertificateCache is a LRU cache to store in memory certificates
+// used for TLS connections
```
iksaif, line 30: Do we really need LRU ? we could just have them expire ?

controllers/datadoghq/tls.go
masterzen: Yes, good idea.

controllers/datadoghq/tls.go
```diff
@@ -0,0 +1,232 @@
...
+type tlsCertificateCache struct {
+	mu    sync.RWMutex
+	cache map[string]*tlsCacheEntry
```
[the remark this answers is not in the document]
masterzen: Not a huge fan of this library. It's not typesafe (agreed that's not a big deal), and it wouldn't even reduce the amount of code compared to my naive implementation because we would still have to deal with the mutex.
If we want to use a lib and agree that we can import a new one, then 
 https://github.com/hashicorp/golang-lru/ is a better choice.

controllers/datadoghq/tls.go
masterzen: It's not a real LRU as it's not memory bounded and entries are not expired until the certificate itself is used and we noticed that it expired. My comment is a lie :)
We can garbage collect the content and remove expired certificates, is that what you suggest?

controllers/datadoghq/tls.go
```diff
@@ -0,0 +1,232 @@
...
+
+// tlsCertificateCache is an expiring cache to store certificates in memory
+// used for TLS connections
+type tlsCertificateCache struct {
+	mu    sync.RWMutex
+	cache map[string]*tlsCacheEntry
+}
+
+// tlsCacheEntry is an entry of the certificate cache
+type tlsCacheEntry struct {
```
vboulineau, line 38: The types and method types are all mixed. For readability, it'd be better to have:
```
type

type methods

type

type methods
```

Free function at the end except `new` right after `type`

bundle/manifests/datadoghq.com_watermarkpodautoscalers.yaml
```diff
@@ -316,6 +316,36 @@ spec:
...
+                        description: Path to the client private key to use for client cert authentication to the server
+                      serverName:
+                        type: string
+                        description: ServerName extension to indicate the name of the server
+                      insecureSkipVerify:
+                        type: string
+                        description: Disable validation of the server certificate
+                      minVersion:
+                        type: string
+                        pattern: ^TLS1[0-3]$
```
vboulineau, line 339: Did you modify the bundle manually? These values are not resent in the Go definition.
You can use `+kubebuilder:validation:Enum` or `+kubebuilder:validation:Pattern` to achieve that with the generator.

masterzen: Yes, I did, I completely forgot that those were generated. My bad.
I'm pushing an auto-generated version.

bundle/manifests/watermarkpodautoscaler.clusterserviceversion.yaml
```diff
@@ -43,7 +43,7 @@ metadata:
     capabilities: Basic Install
     operators.operatorframework.io/builder: operator-sdk-v1.23.0
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
-  name: watermarkpodautoscaler.v0.9.0-rc.1
+  name: watermarkpodautoscaler.v0.9.0-rc.6-497b0e1
```
vboulineau, line 46: Should not be committed

config/manager/kustomization.yaml
```diff
@@ -3,6 +3,6 @@ resources:
 images:
 - name: gcr.io/datadoghq/watermarkpodautoscaler
   newName: gcr.io/datadoghq/watermarkpodautoscaler
-  newTag: v0.9.0-rc.1
+  newTag: v0.9.0-rc.6-497b0e1
```
vboulineau, line 6: Same here

vboulineau (dismissed):

vboulineau (approved):

PR #266, opened 2025-01-14
recommender: adjust metric display

iksaif:
- Make sure it works with `kubectl wpa`
- Use the correct value for the prometheus metric


masterzen (approved):
LGTM!

PR #267, opened 2025-01-15
Update example chart with new CRD definitions

masterzen:
### What does this PR do?

This also fixes the CRD that are in the chart.

### Motivation

In commit cd5a4925, I forgot to update the alongside chart provided with WPA.

### Additional Notes

It's not clear if this is needed because the chart depends on datadog CRD which I have no idea how they are built.
Also not clear if the chart version needs to be bumped (there's a discrepancy between the changelog and the chart version.

### Describe your test plan

This was tested by modifying directly the CRD in an experimental cluster.

### DIFF
--- a/chart/watermarkpodautoscaler/CHANGELOG.md
+++ b/chart/watermarkpodautoscaler/CHANGELOG.md
@@ -1,5 +1,9 @@
 # WatermarkPodAutoscaler Helm chart changelog
 
+## v0.5.1
+
+* CustomResourceDefinition update to fix display of high/low watermarks in some specific cases
+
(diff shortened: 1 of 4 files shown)


iksaif (approved):

fanny-jiang (changes requested):
I've updated the chart version since it wasn't bumped the last time the chart was updated. Can you please bump the chart version for this PR?

masterzen:
@fanny-jiang it should be good now. I was wondering why the chart version hadn't been incremented.
Thanks!

fanny-jiang (approved):

PR #276, opened 2025-03-03
recommender: relay the actual error body

iksaif:
This reads the response body when there's an error and appends it to the error

### DIFF
--- a/controllers/datadoghq/recommender.go
+++ b/controllers/datadoghq/recommender.go
@@ -197,7 +197,16 @@ func (r *RecommenderClientImpl) GetReplicaRecommendation(ctx context.Context, re
 	}
 
 	if resp.StatusCode != http.StatusOK {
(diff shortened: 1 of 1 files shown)


controllers/datadoghq/recommender.go
```diff
@@ -197,6 +197,14 @@ func (r *RecommenderClientImpl) GetReplicaRecommendation(ctx context.Context, re
 	}
 
 	if resp.StatusCode != http.StatusOK {
+		var errorBody []byte
+		if resp.Body != nil {
+			errorBody, _ = io.ReadAll(resp.Body)
```
L3n41c, line 202:
```suggestion
		if resp.Body != nil {
			errorBody, _ := io.ReadAll(resp.Body)
```

L3n41c (dismissed):

justin-lesko (approved):

L3n41c (approved):

PR #278, opened 2025-03-10
Fix Evicted Pods Skewing Available Replicas Calculation

justin-lesko:
### What does this PR do?

The PR fixes a bug related to the `minAvailableReplicaPercentage` option and evicted pods.

> **minAvailableReplicaPercentage:** Indicates the minimum percentage of replicas that need to be available in order for the controller to autoscale the target. For instance, if set at 50 and less than half of the pods behind the target are in an Available state, the target will not be scaled by the controller.

#### Background
According to the [Kubernetes Documention](https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/)
> During a node-pressure eviction, the kubelet sets the [phase](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase) for the selected pods to Failed, and terminates the Pod. 

Because evicted pods remain in the `Terminated` state until garbage collected, they were skewing the `minAvailableReplicaPercentage` calculation because the WPA considers any pod that's been Ready at least once to be part of its target. 

#### Example
```
NAME                                                    READY   STATUS    RESTARTS   AGE
nginx-7b87cc957f-2ws2p                                  0/2     Error     1          37m
nginx-7b87cc957f-4cxmq                                  2/2     Running   0          57s
```
Given the above scenario, previously, the available replica percentage would have been `50%`. Now, with the `Failed` pod phase filter in place, it would be `100%`.
  
### Motivation

A bug report.

### Additional Notes

### Describe your test plan
I first tried simulating evictions by calling the eviction API, but the evicted pods were not persisting as expected. Upon closer inspection of the documentation I learned:
> Node-pressure eviction is not the same as [API-initiated eviction](https://kubernetes.io/docs/concepts/scheduling-eviction/api-eviction/).

From there, I moved on to simulating a `Node-pressure` eviction.

#### Step 1: Deploy Cluster With Limited Resources
I deployed a `k3s` cluster with the following command:
```
k3d cluster create \
  --agents 2 \
  --agents-memory 3g \
  --k3s-arg "--disable=traefik@server:0" \
  --k3s-arg "--kubelet-arg=eviction-hard=memory.available<1Gi@agent:0" \
  --k3s-arg "--kubelet-arg=eviction-hard=memory.available<1Gi@agent:1" \
  --k3s-node-label "role=worker@agent:0,1" 
```
The included flags translate to:
* Deploy 2 "agent" (i.e. worker) nodes
* Agent nodes have 3Gi of memory, overridden from using all of the memory allocated to docker on my local machine
* Both agent nodes' kubelets enforce a hard 1Gi of available memory, and will evict pods that infringe on that
* Agent nodes are tagged with `role=worker`

#### Step 2: Force Memory Consumption
With a low bar for memory eviction in place, I needed a way for my deployed pods to consume enough memory to trigger an eviction. I looked into using `stress-ng` but I wasn't able to find a way to gradually increase memory usage. Ultimately I settled on a small Go program that allocates memory in a loop below.

<details>
<summary>go.mod</summary>
<br>

```
module memory-eater

go 1.23
```
</details>

<details>
<summary>main.go</summary>
<br>

```
package main

    import (
      "fmt"
      "runtime"
      "runtime/debug"
      "time"
    )

    func main() {
      // turn off garbage collection
      debug.SetGCPercent(-1)

      var memUsage []byte

      // allocate 5MB chunks
      allocSize := 5 * 1024 * 1024 

      fmt.Println("Starting memory consumption...")

      for i := 1; ; i++ {
        block := make([]byte, allocSize)
        for j := range block {
          block[j] = 1 // Actively write to memory to keep the memory in use
        }
        memUsage = append(memUsage, block...)

        memStats := &runtime.MemStats{}
        runtime.ReadMemStats(memStats)
        fmt.Printf("Iteration: %d | Allocated Memory: %.2f MB\n", i, float64(memStats.Alloc)/1024/1024)

        time.Sleep(1 * time.Second)
      }
    }

```
</details>

#### Step 3: Deploy a WPA Scalable Deployment
I deployed a basic Nginx deployment with a few key modifications
1. A taint toleration, so pods can still be scheduled after the node experiences memory pressure
    ```
    tolerations:
      - key: "node.kubernetes.io/memory-pressure"
        operator: "Exists"
        effect: "NoSchedule"
    ```
2. A node selector to target the agent nodes
    ```
    nodeSelector:
        role: worker
    ```
3. A sidecar container with the memory-eater application mounted
    ```
    - name: memory-eater
      image: golang:1.23
      imagePullPolicy: Always
      workingDir: /usr/src/memory-eater
      command: ["/bin/sh", "-c", "sleep infinity"]
      volumeMounts:
        - name: go-program
          mountPath: /usr/src/memory-eater
    ```

#### Step 4: Deploy/Configure the WPA
Deploy the WPA with the `minAvailableReplicaPercentage` option 

```
apiVersion: datadoghq.com/v1alpha1
kind: WatermarkPodAutoscaler
metadata:
  name: nginx
spec:
  algorithm: average
  minReplicas: 3
  maxReplicas: 30
  tolerance: "0.0"
  minAvailableReplicaPercentage: 50 # <--This is key to the test
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: nginx
  metrics:
  - type: External
    external:
      highWatermark: 9
      lowWatermark: 8
      metricName: datadogmetric@<NAMESPACE>:nginx
      metricSelector: {}
```

#### Step 5: Trigger evictions
Trigger enough evictions so that evicted pods outnumber ready pods by running the following command N times:
```
# Issue an exec command without in the background, without tailing the logs
kubectl exec <POD> -c memory-eater -n <NAMESPACE> -- nohup go run . >/dev/null 2>&1 &
```

#### Step 6: Generate Load
I port-forwarded to my Nginx service

```
kubectl port-forward service/nginx -n workload-nginx 8080:80
```

Then I used k6 to generate a constant amount of load to trigger a desired scale-up

```
import http from 'k6/http';
import { check } from 'k6';

export let options = {
  scenarios: {
    fixed_rps: {
      executor: 'constant-arrival-rate',
      rate: 50, 
      timeUnit: '1s',
      duration: '30m',
      preAllocatedVUs: 100,
      maxVUs: 200,
    }
  }
};

export default function () {
  const res = http.get('http://localhost:8080');
  check(res, { 'status was 200': (r) => r.status == 200 }); // probably unnecessary
}
```

##### Old Behavior- stuck, unable to scale while evicted pods skew available replica calculation
WPA Failure Behavior

##### New Behavior- able to scale despite the evicted pods persisting
WPA Fix Behavior

### DIFF
--- a/controllers/datadoghq/replica_calculator.go
+++ b/controllers/datadoghq/replica_calculator.go
@@ -113,7 +113,6 @@ func (c *ReplicaCalculator) GetExternalMetricReplicas(logger logr.Logger, target
 	}
 
 	ratioReadyPods := (100 * currentReadyReplicas / (int32(len(podList)) - incorrectTargetPodsCount))
-
 	if ratioReadyPods < wpa.Spec.MinAvailableReplicaPercentage {
 		return ReplicaCalculation{}, fmt.Errorf("%d %% of the pods are unready, will not autoscale %s/%s", ratioReadyPods, target.Namespace, target.Name)
 	}
--- a/controllers/datadoghq/replica_calculator.go
+++ b/controllers/datadoghq/replica_calculator.go
@@ -517,10 +517,17 @@ func (c *ReplicaCalculator) getReadyPodsCount(log logr.Logger, targetName string
 			continue
 		}
 
-		// During a node-pressure eviction, the kubelet sets the phase for the selected pods to Failed, and terminates the Pod.
+		// PodFailed
+		//During a node-pressure eviction, the kubelet sets the phase for the selected pods to Failed, and terminates
+		// the Pod. These pods should be ignored because they may not be garbage collected for a long time.
 		// https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/
-		// These pods should be ignored because they may not be garbage collected for a long time
-		if pod.Status.Phase == corev1.PodFailed {
+
+		// PodSucceeded
+		// A Deployment’s Pod should never be in PodSucceeded. If it is, it usually means:
+		// - The Pod is running a one-shot script instead of a service.
+		// - The restartPolicy is misconfigured.
+		// - A Job-like process was accidentally set up in a Deployment.
+		if pod.Status.Phase == corev1.PodFailed || pod.Status.Phase == corev1.PodSucceeded {
 			incorrectTargetPodsCount++
 			continue
 		}
--- a/controllers/datadoghq/replica_calculator.go
+++ b/controllers/datadoghq/replica_calculator.go
@@ -113,6 +113,7 @@ func (c *ReplicaCalculator) GetExternalMetricReplicas(logger logr.Logger, target
 	}
 
 	ratioReadyPods := (100 * currentReadyReplicas / (int32(len(podList)) - incorrectTargetPodsCount))
+
 	if ratioReadyPods < wpa.Spec.MinAvailableReplicaPercentage {
 		return ReplicaCalculation{}, fmt.Errorf("%d %% of the pods are unready, will not autoscale %s/%s", ratioReadyPods, target.Namespace, target.Name)
 	}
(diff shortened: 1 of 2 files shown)


vboulineau (approved):

controllers/datadoghq/replica_calculator.go
```diff
@@ -516,6 +516,15 @@ func (c *ReplicaCalculator) getReadyPodsCount(log logr.Logger, targetName string
 			incorrectTargetPodsCount++
 			continue
 		}
+
+		// During a node-pressure eviction, the kubelet sets the phase for the selected pods to Failed, and terminates the Pod.
+		// https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/
+		// These pods should be ignored because they may not be garbage collected for a long time
+		if pod.Status.Phase == corev1.PodFailed {
```
clamoriniere, line 523: I think I saw also `corev1.PodSucceeded` occurence in the path. that was not cleanup fast enough.
For what I understand `PodSucceeded` is more for Job's pod. but I feel it could happen also with eviction. 

can we also ignore them?

justin-lesko: That sounds like a good improvement, thanks!

clamoriniere (approved):

PR #285, opened 2025-03-17
Add more metrics to detect broken WPAs

iksaif:
### What does this PR do?

This adds:
- per WPA `scaling_active` that reflects the value of the associated condition
- per HTTP client `http_client_requests_errors_total` to count client errors (and not http errors)

### Motivation

Detect WPAs that might have trouble contacting their recommenders


iksaif:
Tested on lagaffe

vboulineau (approved):

PR #359, opened 2026-07-28
Onboard off public-images GitLab job-token trigger

AliDatadog:
Replaces the legacy `trigger: project: DataDog/public-images` job with a direct `dd-pkg publish-image` call against artifact-gateway.

## Why

The GitLab job-token bridge cannot tell **who** triggered a publish — any project holding a token could publish anything. Calling artifact-gateway instead adds:

- **Authentication** — the caller's identity is verified.
- **Authorization** — a policy states which Directory Service group may release each image tier (dev vs prod).
- An audit trail of every publication.

**Authorization is in audit-only mode for now** — nothing is blocked, publications are only recorded. So this is safe to merge, and the rollout stays reversible behind a feature flag.

Publish cadence, tags and destinations are unchanged; this is an onboarding change only.

- Jira: [BARX-1952](https://datadoghq.atlassian.net/browse/BARX-1952)
- How-to: [Publish public images with dd-pkg](https://datadoghq.atlassian.net/wiki/x/ooQkngE)

[BARX-1952]: https://datadoghq.atlassian.net/browse/BARX-1952?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -79,6 +79,7 @@ build_image_fips:
 
 .docker_publish_job_definition:
   image: registry.ddbuild.io/agent-delivery/dd-pkg:v0.9.0
+  tags: ["arch:amd64"]
   before_script: []
   variables:
     IMG_SIGNING: "false"
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -79,7 +79,7 @@ build_image_fips:
 
 .docker_publish_job_definition:
   image: registry.ddbuild.io/agent-delivery/dd-pkg:v0.9.0
-  tags: ["arch:amd64"]
+  tags: ["arch:arm64"]
   before_script: []
   variables:
     IMG_SIGNING: "false"
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -79,8 +79,10 @@ build_image_fips:
 
 .docker_publish_job_definition:
   image: registry.ddbuild.io/agent-delivery/dd-pkg:v0.9.0
+  before_script: []
   variables:
     IMG_SIGNING: "false"
+    IMG_REGISTRIES: "public"
     PUBLIC_IMAGES_PUBLISH_TIMEOUT: "1800"
   script:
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -77,16 +77,32 @@ build_image_fips:
     RELEASE_IMAGE: $BUILD_DOCKER_REGISTRY/$PROJECTNAME:$CI_COMMIT_TAG-fips
     FIPS_ENABLED: "true"
 
+.docker_publish_job_definition:
+  image: registry.ddbuild.io/agent-delivery/dd-pkg:v0.9.0
```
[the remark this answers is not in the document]
AliDatadog: Good catch — fixed. The publish template now sets `before_script: []`, so the converted jobs no longer inherit the top-level `before_script` and no longer run `make install-tools` inside the `dd-pkg` image.

.gitlab-ci.yml
```diff
@@ -77,16 +77,35 @@ build_image_fips:
     RELEASE_IMAGE: $BUILD_DOCKER_REGISTRY/$PROJECTNAME:$CI_COMMIT_TAG-fips
     FIPS_ENABLED: "true"
 
+.docker_publish_job_definition:
+  image: registry.ddbuild.io/agent-delivery/dd-pkg:v0.9.3
+  tags: ["arch:arm64"]
+  before_script: []
```
L3n41c, line 83: Can the `before_script` section be completely removed or does it need to be explicitly set to an empty list?


```suggestion
```

L3n41c (approved):

.gitlab-ci.yml
AliDatadog: It needs to be explicitely set to an empty list because there is a top-level before_script running `make install-tools`. Since the image doesn't have make or go, it would fail. Removing this would require some refactoring

PR #367, opened 2026-08-26
[CASCL-1709] Fix stuck WPA restricted_scaling capping metric

clamoriniere:
### What does this PR do?

Fixes the `watermarkpodautoscaler.wpa_controller_restricted_scaling{reason:upscale_capping}` (and `downscale_capping`) gauge staying "stuck" at `1` after scaling was capped once. `convertDesiredReplicasWithRules` now deterministically refreshes both the `upscale_capping` and `downscale_capping` gauges on every return path, so a gauge set to `1` during one reconcile is reset to `0` on subsequent reconciles that no longer cap in that direction.

### Motivation

Addresses Jira CASCL-1709. During an incident, a user observed that `upscale_capping` kept reporting `1` long after the proposed replica count dropped below the effective replica count — it only reset on a metric-fetch error or controller restart. Root cause is in `convertDesiredReplicasWithRules`:

- The shared label map was initialized to `downscale_capping` and only switched to `upscale_capping` inside the "capped" branch. On the non-capping path the code called `Set(0)` while the label was still `downscale_capping`, so `upscale_capping` was never reset.
- The early `return` on the downscale-capping / below-minimum path skipped the upscale block entirely, leaving `upscale_capping` untouched.
- The `wpaMinReplicas == 0` switch case never touched `downscale_capping`.

`normalizeDesiredReplicas` → `convertDesiredReplicasWithRules` runs on every successful reconcile, so correcting this function is sufficient to keep the gauges accurate.

### Changes

- `controllers/datadoghq/watermarkpodautoscaler_controller.go`: track capping decisions with local `downscaleCapping` / `upscaleCapping` booleans and set both gauges from a single `defer`, each with its own freshly-built label map. Replaced the branch-local `restrictedScaling.With(...).Set(...)` calls with flag assignments. Scaling behavior (returned replicas, conditions, reasons) is unchanged; only metric bookkeeping is corrected.
- `controllers/datadoghq/metrics.go`: add a small `boolToFloat64` helper for the 1/0 gauge representation.
- `controllers/datadoghq/watermarkpodautoscaler_controller_test.go`: assert the `upscale_capping` / `downscale_capping` gauge values in `TestConvertDesiredReplicasWithRules` (seeding both to `1` beforehand to prove they are refreshed), and add `TestConvertDesiredReplicasWithRulesResetsRestrictedScaling`, a regression test that reproduces the CASCL-1709 sequence (cap → within-range reset) for both directions. Float gauge assertions use `require.InDelta` (matching the existing convention in this file) to satisfy the `testifylint` `float-compare` linter.

### Additional Notes

No API/CRD changes, so no `make generate`/`manifests` needed. On metric-fetch errors the replica calculator already deletes these gauges, which is consistent with the previously reported "resets on error" behavior.

### Describe your test plan

- `go test ./controllers/datadoghq/` — full package suite passes.
- `go test ./controllers/datadoghq/ -run 'TestConvertDesiredReplicasWithRules' -v` — new and updated assertions pass; the regression test fails against the old code and passes with the fix.
- `testifylint ./controllers/datadoghq/` — clean (the float gauge assertions now use `require.InDelta`, resolving the CI lint failures).
- `go vet ./controllers/datadoghq/` and `gofmt -l` are clean.

---

PR by Bits - [View session in Datadog](https://app.datadoghq.com/code/a1524404-5a8a-45de-90f2-6a39ab236136)

Comment @datadog to request changes

### DIFF
--- a/controllers/datadoghq/watermarkpodautoscaler_controller.go
+++ b/controllers/datadoghq/watermarkpodautoscaler_controller.go
@@ -949,26 +949,19 @@ func normalizeDesiredReplicas(logger logr.Logger, wpa *datadoghqv1alpha1.Waterma
 func convertDesiredReplicasWithRules(logger logr.Logger, wpa *datadoghqv1alpha1.WatermarkPodAutoscaler, currentReplicas, desiredReplicas, wpaMinReplicas, wpaMaxReplicas int32, readyReplicas int32) (int32, string, string) {
 	var minimumAllowedReplicas int32
 	var maximumAllowedReplicas int32
+	var normalizedReplicas int32
 	var possibleLimitingCondition string
 	var possibleLimitingReason string
 
 	// Track the capping decisions so that both the downscale_capping and
-	// upscale_capping restricted_scaling gauges are refreshed on every return
-	// path. Previously these gauges were only updated on the specific branch
-	// that computed them, so a gauge set to 1 during one reconcile could stay
-	// "stuck" at 1 on subsequent reconciles that no longer capped in that
-	// direction (see CASCL-1709).
+	// upscale_capping restricted_scaling gauges are refreshed on every
+	// reconcile. Previously these gauges were only updated on the specific
+	// branch that computed them, so a gauge set to 1 during one reconcile could
+	// stay "stuck" at 1 on subsequent reconciles that no longer capped in that
+	// direction (see CASCL-1709). The function funnels every path through a
+	// single exit so both gauges are always refreshed before returning.
 	downscaleCapping := false
 	upscaleCapping := false
-	defer func() {
-		downscaleLabels := getPrometheusLabels(wpa)
-		downscaleLabels[reasonPromLabel] = downscaleCappingPromLabelVal
-		restrictedScaling.With(downscaleLabels).Set(boolToFloat64(downscaleCapping))
-
-		upscaleLabels := getPrometheusLabels(wpa)
-		upscaleLabels[reasonPromLabel] = upscaleCappingPromLabelVal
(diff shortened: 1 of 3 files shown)


triviajon (dismissed):
resetting the reason labels on the gauges on every reconciliation and removing the early exits LGTM

controllers/datadoghq/watermarkpodautoscaler_controller.go
```diff
@@ -949,20 +949,29 @@ func normalizeDesiredReplicas(logger logr.Logger, wpa *datadoghqv1alpha1.Waterma
...
 	var possibleLimitingCondition string
 	var possibleLimitingReason string
 
+	// Track the capping decisions so that both the downscale_capping and
+	// upscale_capping restricted_scaling gauges are refreshed on every
+	// reconcile. Previously these gauges were only updated on the specific
+	// branch that computed them, so a gauge set to 1 during one reconcile could
+	// stay "stuck" at 1 on subsequent reconciles that no longer capped in that
+	// direction (see CASCL-1709). The function funnels every path through a
+	// single exit so both gauges are always refreshed before returning.
```
triviajon: nit: small but this explanation reappears again in `controllers/datadoghq/watermarkpodautoscaler_controller.go:1016-1017` and `controllers/datadoghq/watermarkpodautoscaler_controller_test.go:2444-2448`, just the one is probably fine

clamoriniere:
```suggestion
```

triviajon (approved):

PR #374, opened 2026-09-15
CASCL-1693: Migrate off autoscaling/v2beta1 and v2beta2, bump k8s.io/api to v0.36.0

clamoriniere:
## Summary

Kubernetes 1.36 drops the `autoscaling/v2beta1` and `autoscaling/v2beta2` API versions, and `k8s.io/api` v0.36.0 removes the corresponding Go packages entirely (confirmed by inspecting the module directly), so the controller could no longer compile against that dependency version as-is.

- Localizes the `WatermarkPodAutoscaler` status types whose JSON shape actually changed between `v2beta1` and `v2` (`MetricStatus`/`ExternalMetricStatus`/`ResourceMetricStatus`) into `apis/datadoghq/v1alpha1`, keeping the exact same wire format so existing WPA consumers (kubectl printcolumns, dashboards, scripts) are unaffected. `HorizontalPodAutoscalerCondition` and its consts are unchanged in shape between `v2beta1` and `v2`, so those are read directly from `k8s.io/api/autoscaling/v2` rather than duplicated.
- Bumps `k8s.io/api`, `apimachinery`, `client-go`, `apiextensions-apiserver` and `controller-runtime` (v0.21.0 -> v0.24.0, matching the version that itself pins `k8s.io/api` at v0.36.0) in both `go.mod` and the `apis` submodule's `go.mod`.
- Follow-up: migrates the controller from the deprecated `record.EventRecorder`/`GetEventRecorderFor` to the new `events.EventRecorder`/`GetEventRecorder` API (required by the controller-runtime v0.24.0 deprecation), reusing existing `reason` values as the new required `action` field, and wiring the `DatadogMonitor` object as the `related` object for the lifecycle-control events that are gated on it.
- Updates RBAC from `groups=\"\",resources=events` to `groups=events.k8s.io,resources=events` to match the new events API.

## Minimum Kubernetes version

This bumps the minimum compatible Kubernetes version to **1.19**, solely because event recording now posts `events.k8s.io/v1` Events (via the new `client-go` events API / `mgr.GetEventRecorder()`) instead of legacy `core/v1` Events. `events.k8s.io/v1` graduated to GA in Kubernetes 1.19; on older clusters event recording would silently no-op (best-effort, non-blocking) but scaling logic itself would be unaffected.

Nothing else in this PR raises the floor further: the `autoscaling/v2` Go types reused for the WPA's own CRD status are never sent to the real `autoscaling/v2` HPA API (WPA doesn't create/read HPA objects), so there's no dependency on that API group being served. The other APIs this PR touches - the `Scale` subresource, `apiextensions.k8s.io/v1` CRDs, `rbac.authorization.k8s.io/v1` - are unchanged and have been stable since much older releases.

## e2e CI infrastructure

The e2e suite talks to a fake `external.metrics.k8s.io/v1beta1` adapter to exercise WPA's external-metrics path. Getting it working against the new client-go/API versions required:

- Vendoring the fake metrics-server adapter into the repo (`test/e2e/metricsserver/adapter`) and building it locally instead of relying on a prebuilt external image
- Waiting for the fake external metrics APIService to report `Available` before running scaling assertions, instead of racing it
- Extra diagnostics (APIService status, aggregated discovery state) logged on failure to make future CI flakes debuggable
- Scoping `controller-gen` to explicit source trees instead of `./...` (it was picking up the vendored adapter module)

### e2e Kubernetes version matrix

Also updated the e2e job's `kind` node-image matrix, since it was badly out of date and didn't reflect the versions this PR now targets:

| Before | After |
|---|---|
| v1.14.10 | v1.19.16 *(new minimum, see above)* |
| v1.16.9 | v1.25.16 |
| v1.18.4 | v1.31.14 |
| v1.21.1 | v1.36.4 *(latest)* |

The old matrix topped out at 1.21.1 (released 2021) and never tracked further - there was a growing gap between what CI actually exercised and the Kubernetes versions users run in production today, including the current latest release. The new matrix spans from the new minimum (1.19) up to the current latest (1.36), with two representative versions in between, so CI coverage now matches the versions we claim to support.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...` and `go vet -tags e2e ./...`
- [x] `make build` (fmt, vet, `golangci-lint` - 0 issues, binaries)
- [x] `go test ./...` (including envtest-backed controller tests)
- [x] `make generate manifests` and reviewed the resulting CRD/RBAC diffs for backward compatibility (`status.conditions` unchanged; `status.currentMetrics[*].external/resource` keeps identical field names, only drops never-populated sub-schemas)
- [x] `make e2e` / `make goe2e` against a real cluster

🤖 Generated with [Claude Code](https://claude.com/claude-code)


test/e2e/metricsserver/adapter/provider/provider.go
```diff
@@ -0,0 +1,434 @@
+/*
```
vboulineau, line 1: Not sure we needed to all this whole package there?

clamoriniere: This is the code for the fake metrics service that we use in place of the Cluster Agent. It is based on the Kubernetes [custom metrics test adapter](https://github.com/kubernetes-sigs/custom-metrics-apiserver/tree/master/test-adapter), but we have historically maintained a custom build that uses a ConfigMap as its metrics store.

After I updated the Kubernetes version used in CI, the existing fake metrics server stopped working with newer clusters. I copied the adapter code here and updated it so we can rebuild the server with recent dependencies.
The package remains small and straightforward. I could remove the functionality that our end-to-end tests don’t use, but I don’t think that would provide much value and it would make the code diverge further from upstream.

vboulineau (approved):

